PR #2, opened 2023-03-15
Enhancing the extract-auth-token.py script

thomasjpatterson:
Added cli flags of -h (--help), -f (--file), and -a (--auto). Help displays correct syntax. Auto will attempt to detect your OS and then execute the script. File you will need to put in the correct path for it to correctly grab your auth token.

### DIFF
--- a/examples/extract-auth-token.py
+++ b/examples/extract-auth-token.py
@@ -6,27 +6,43 @@ import json
 import sys
 import platform
 import os
+import getopt
 
-userdir = os.getlogin()
-useros = platform.system()
+def jankauto():
+    userdir = os.getlogin()
+    useros = platform.system()
 
-#darfileloc = f'/Users/{userdir}/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json'
-darfileloc = f'/Users/thomaspatterson/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json'
-winfileloc = os.path.join('C:\\Users\\') + f'{userdir}' + os.path.join('\\AppData\\Local\\Ankermake\\login.json')
+    #darfileloc = f'/Users/{userdir}/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json'
+    darfileloc = f'/Users/thomaspatterson/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json'
+    winfileloc = os.path.join('C:\\Users\\') + f'{userdir}' + os.path.join('\\AppData\\Local\\Ankermake\\login.json')
+    
+    try:
+        opts, args = getopt.getopt(sys.argv[1:],"haf:", ["help", "auto", "file="])
+    except getopt.GetoptError:
+      print ('extract-auth-token.py -a OR extract-auth-token.py -f <path-to-login.json>')
+      sys.exit(2)
 
-if useros == 'Darwin':
-    data = open(f'{darfileloc}').read()
-    json = json.loads(libflagship.logincache.decrypt(data))
-    print(json["data"]["auth_token"])
-    exit(1)
-elif useros == 'Windows':
-    data = open(f'{winfileloc}').read()
-    json = json.loads(libflagship.logincache.decrypt(data))
-    print(json["data"]["auth_token"])
-    exit(1)
-elif len(sys.argv) != 2:
-    print(f"usage: {sys.argv[0]} <path/to/login.json>")
-    data = open(sys.argv[1]).read()
-    json = json.loads(libflagship.logincache.decrypt(data))
-    print(json["data"]["auth_token"])
-    exit(1)
+    for opt, arg in opts: 
+        if opt == '-h':
+            print ('extract-auth-token.py -a OR extract-auth-token.py -f <path-to-login.json> \nIf spaces are in file location remember to escape them with a backslash')
+            exit(1)
+        elif opt in ("-f", "--file"):
+            data = open(arg).read()
+            jsonj = json.loads(libflagship.logincache.decrypt(data))
+            print(jsonj["data"]["auth_token"])
+            exit(1)
+        elif opt in ("-a", "--auto"):
+            if useros == 'Darwin':
+                data = open(f'{darfileloc}').read()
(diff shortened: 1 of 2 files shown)


examples/extract-auth-token.py
```diff
@@ -4,12 +4,44 @@
...
+                exit(1)
+            elif useros == 'Windows':
+                data = open(f'{winfileloc}').read()
+                jsonj = json.loads(libflagship.logincache.decrypt(data))
+                print(jsonj["data"]["auth_token"])
+                exit(1)
+        else:
+            assert False, "unhandled option"
+
+jankauto()
```
chrivers, line 47: If we're wrapping "everything" in a function, a standardish python idiom is something like:

```python
if __name__ == "__main__":
    exit(main())
```

at the end of the file. That way, the `main()` function can just return whatever exit code is appropriate.

examples/extract-auth-token.py
```diff
@@ -4,12 +4,44 @@
...
-if len(sys.argv) != 2:
-    print(f"usage: {sys.argv[0]} <path/to/login.json>")
-    exit(1)
+def jankauto():
+    userdir = os.getlogin()
+    useros = platform.system()
 
-data = open(sys.argv[1]).read()
+    darfileloc = f'/Users/{userdir}/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json'
+    winfileloc = os.path.join('C:\\Users\\') + f'{userdir}' + os.path.join('\\AppData\\Local\\Ankermake\\login.json')
```
chrivers, line 16: We can do this slightly cleaner. On windows, we can use `%APPDATA%` to refer specifically to the user's appdata dir:

```python
from os import path
...
win_path = path.expandvars(r'%LOCALAPPDATA%\Ankermake\login.json')
# or (possibly more compatible, not sure)
win_path = path.expandvars(r'%APPDATA%\Local\Ankermake\login.json')
```

For mac, we could do:

```python
mac_path = path.expandhome("~/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json")
```

These are built into python, but maybe something like [platformdirs](https://pypi.org/project/platformdirs/) might be even smoother?

examples/extract-auth-token.py
```diff
@@ -4,12 +4,44 @@
...
+    
+    try:
+        opts, args = getopt.getopt(sys.argv[1:],"haf:", ["help", "auto", "file="])
+    except getopt.GetoptError:
+      print ('extract-auth-token.py -a OR extract-auth-token.py -f <path-to-login.json>')
+      sys.exit(2)
```
chrivers, line 22: If you feel up for it, consider using `argparse`, for an easier, more modern argument parsing (see `connect-mqtt.py` for an example)

See the last comment for how to wrap `main()`, then replace `sys.exit(2)` with just `return 2` :)

examples/extract-auth-token.py
```diff
@@ -4,12 +4,44 @@
...
+            jsonj = json.loads(libflagship.logincache.decrypt(data))
+            print(jsonj["data"]["auth_token"])
+            exit(1)
+        elif opt in ("-a", "--auto"):
+            if useros == 'Darwin':
+                data = open(f'{darfileloc}').read()
+                jsonj = json.loads(libflagship.logincache.decrypt(data))
+                print(jsonj["data"]["auth_token"])
+                exit(1)
+            elif useros == 'Windows':
```
chrivers, line 39: There's a lot of repeated code here. Please extract this repeated pattern into a function:

```python
def print_login(filename):
    data = open(filename).read()
    jsonj = json.loads(libflagship.logincache.decrypt(data))
    print(jsonj["data"]["auth_token"])
```

Then just call it in 3 places.

examples/extract-auth-token.py
```diff
@@ -4,12 +4,44 @@
...
+                data = open(f'{winfileloc}').read()
+                jsonj = json.loads(libflagship.logincache.decrypt(data))
+                print(jsonj["data"]["auth_token"])
+                exit(1)
```
chrivers, line 43: Any non-zero exit status is meant to indicate failure of some kind. We should return 0 (or simply not calling `sys.exit()`) if everything went well.

examples/extract-auth-token.py
```diff
@@ -4,12 +4,44 @@
...
+        else:
+            assert False, "unhandled option"
```
chrivers, line 45: this is more or less automatically handled by `argparse`, but if we stay with `getopt()`, I think `assert` is not the best choice here. It'd be better to print the usage string, and `exit(1)`.

chrivers (changes requested):
Thank you for the PR! I've made some suggestions for improving this in various places. I'm not trying to be a soup nazi about this, but I hope you are up for reworking it a bit. Thanks :)

examples/extract-auth-token.py
```diff
@@ -4,12 +4,43 @@
...
+    print(jsonj["data"]["auth_token"])
 
-data = open(sys.argv[1]).read()
+def main():
+    userdir = os.getlogin()
+    useros = platform.system()
 
-json = json.loads(libflagship.logincache.decrypt(data))
-print(json["data"]["auth_token"])
+    darfileloc = f'/Users/{userdir}/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json'
```
chrivers, line 20: can you test this?

```python
    darfileloc = os.path.expanduser('~/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json')
```

Then we don't need to make assumptions about where homedir is located (which is what we're really trying to do)

examples/extract-auth-token.py
```diff
@@ -4,12 +4,43 @@
...
+    try:
+        opts, args = getopt.getopt(sys.argv[1:],"haf:", ["help", "auto", "file="])
+    except getopt.GetoptError:
+      print ('extract-auth-token.py -a OR extract-auth-token.py -f <path-to-login.json>')
+      return 2
+
+    for opt, arg in opts: 
+        if opt == '-h':
+            print ('extract-auth-token.py -a OR extract-auth-token.py -f <path-to-login.json> \nIf spaces are in file location remember to escape them with a backslash')
+            return(1)
```
chrivers, line 32: `return 1` - return is not a function

PR #5, opened 2023-03-17
+Added MQTT and Example File Documentation

austinrdennis:
Added documentation on the use of the MQTT example file. Added requirements.txt to TLD for use in semi-automated install of pip prerequisites.


austinrdennis:
@chrivers Before final merge, I want to get your input on where you want to add the link to this on the README file.

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+   pip3 install -r requirements.txt
+   ```
+
+   If required, enter `Y` to all installation prompts.
+
+
+
+## Extracting your Auth Token
+
+Your Auth Token is the "key to the kingdom" so-to-speak. The M5 uses this to verify who can connect to it over the internet (or LAN) to make sure it's actually you or someone you've shared your printer with (although, we haven't done any testing with shared printers so that's more of an educated guess). You're printer will refuse any connections that don't present your unique Auth Token, so let's get that out for future use.
```
chrivers, line 75: `s/You're/Your/`

And I have to clear up a technical misunderstanding here. The printer actually doesn't know about the `auth_token`! The token is used purely for a few endpoints in the https api, but we use that api to extract the mqtt key, and a few other key pieces of information.

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. Navigate to wherever you cloned this repository to. Open the "examples" folder and open a terminal window there, just like in the previous section.
+
+2. Type in the following command:
+
+   ```bash
+    python3 extract-auth-token.py -a
+   ```
+
+   You should get back a long sequence of numbers and letters, that's your Auth Token. Copy that to the clipboard for use in the next section.
```
chrivers, line 85: Maybe add an example with `-f /path/to/login.json` ? That would help people with non-standard install locations.

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+
+## Connecting to your M5 over MQTT
+
+Now here comes the fun part, we have everything we need to actually connect to the printer. It uses the [MQTT standard](https://mqtt.org/), to communicate with the wider world. It is possible to use this standard coupled with this repository build a tool that can remotely monitor the printer, print files, execute movement commands, pre-heat the printer, initiate the Auto-Bed Leveling routine, and just about anything else the companion phone app can. In this example, we'll simply be monitoring the printer. We can also save the output to a log file if desired.
```
chrivers, line 91: `It is possible to use this standard coupled with this repository build a tool`

I think this sentence a verb?

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+   python3 mqtt-connect.py -r us -A "YOUR_AUTH_TOKEN_HERE"
+   ```
+
+3. **[Optional Step]** If desired, you can save the contents of the output to a log file by adding `> output.log` to the end of the command in the previous step:
+
+   ```bash
+   python3 mqtt-connect.py -r us -A "YOUR_AUTH_TOKEN_HERE" > output.log
+   ```
+
+Now you should see a bunch of data updating in your terminal, this happens in 1 second intervals. It should look something like this and will be quite a bit more colorful:
```
chrivers, line 107: `s/1 second interval/roughly once per second/`

It can sometimes be a few seconds between updates, and sometimes less than a second.

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+    packet_type=<MqttPktType.Single: 192>,
+    packet_num=1,
+    time=1678924548,
+    device_guid='SOME_DATA_HERE',
+    data=b'[{"commandType":1003,"currentTemp":5263,"targetTemp":18000},{"comman
+dType":1004,"currentTemp":3990,"targetTemp":6000}]'
+)
+```
+
+There's quite a bit to see, but let's just talk about the more interesting data from a layman's perspective  here:
```
chrivers, line 128: double spacing.

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+
+The TOPIC (that giant wall of numbers and letters at the top) is a string of data that represents what the printer is doing right now (homing, pre-heating, probing, etc.). When building your program, you could translate these messages to meaningful status messages for display.
```
chrivers, line 130: Sorry, that's wrong.

The "TOPIC" is the part in brackets:

`/phone/maker/YOUR_SERIAL_NUMBER_HERE/notice`

Think of it as the mqtt "path" that the message happens on. It's essentially a named queue of messages.

The hex string (`4d41c1000501020546c0010....`) is the raw payload bytes.

Decrypting and decoding those bytes yield the structured message (`MqttMsg(size=193, ... `)

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
...
+
+The MQTTMsg is what contains a bunch little data points about the printer like nozzle temp, bed temp, what command it most recently received, etc. When building your program, you could use these messages to update data live as long as the MQTT connection was established.
```
chrivers, line 132: This is technically correct, but maybe a little bit misleading.

The MQTTMsg is a standard encapsulation used by all messages over the mqtt connection.

Almost all the interesting data is in the `.data` field of the MQTTMsg, which is a json-encoded command. So that's essentially the next level of (un)wrapping, which we're just displaying as a text string right now.

chrivers (changes requested):
Really nice! A few minor misunderstandings, and some nitpickery :wink:

documentation/example-file-usage/mqtt-example-file-usage.md
austinrdennis: Ok, I definitely have a technical concept error here then. I'd love some amplifying info on that flow and maybe I can add that as a flow chart here instead.

requirements.txt
```diff
@@ -0,0 +1,4 @@
+paho_mqtt==1.6.1
+pycryptodomex==3.17
+rich==13.3.1
+requests==2.28.2
```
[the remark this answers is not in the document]
austinrdennis: pip doesn't like this. I get this output:

`Collecting https://github.com/chrivers/transwarp.git (from -r requirements.txt (line 5))
  Downloading https://github.com/chrivers/transwarp.git
     | 230.1 kB 1.6 MB/s 0:00:00
  ERROR: Cannot unpack file C:\Users\austi\AppData\Local\Temp\pip-unpack-19gogmdf\transwarp.git (downloaded from C:\Users\austi\AppData\Local\Temp\pip-req-build-zamqlhgy, content-type: text/html; charset=utf-8); cannot detect archive format
ERROR: Cannot determine archive format of C:\Users\austi\AppData\Local\Temp\pip-req-build-zamqlhgy`

Any idea what I should try?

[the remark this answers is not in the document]
chrivers: Ah, it seems the syntax is a bit different:

`git+https://...`

documentation/example-file-usage/mqtt-example-file-usage.md
```diff
@@ -0,0 +1,136 @@
+# How to connect to a M5 over CLI (via MQTT)
+
+We've prepared a couple example files to demonstrate the functionality of this repository and it's intended to be used mostly by developers, but the purpose of this guide is to get the average user to the point where they can run the example files on their machine and actually interface with the printer.
```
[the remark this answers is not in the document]
austinrdennis: On a second read, a lot of this seems like unnecessary fluff. I just cut a lot of it out as it really doesn't belong in a guide like this.

documentation/example-file-usage/mqtt-example-file-usage.md
chrivers: Sure thing!

When the printer is turned on, it will start sending regular mqtt updates to Ankermake mqtt servers. The script we're running here, will connect to those same servers, and subscribe to the same channels.

That way, we end up seeing the printer messages, but they're all bounced through Anker's servers. That's also how the slicer does it.

In fact, it's the only way to do it - both the printer and our script work as mqtt clients, so they need to connect to some mqtt server out there. Since the printer always connect to the ankermake servers, we do the same thing.

chrivers: And that's (indirectly) why we need the `auth_token`: to be able to connect to the https api, and request the mqtt login information.

requirements.txt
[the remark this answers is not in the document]
austinrdennis: Perfect! It works on at least Windows and I imagine it will on all the other platforms.

austinrdennis:
Re-worked and ready for more nitpicks 😉

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
+# AnkerMake MQTT Communications Overview
+
+We're going over the AnkerMake implementation of the [MQTT communications standard](https://mqtt.org/). While we won't be explaining the MQTT standard itself, we will explain certain aspects of how the protocol works for ease of understanding.
```
chrivers, line 3: Anker(Make) didn't implement MQTT - they're using bog standard client and server implementations.

However, they did so some fresh jazz to the bytes going over the protocol.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Generic MQTT Basics
+
+MQTT (Messaging Que Transport Telemetry) is a [stateful](https://byjus.com/gate/difference-between-stateless-and-stateful-protocol/#:~:text=What%20is%20a%20Stateful%20Protocol,Protocol%20(FTP)%2C%20etc.), [edge driven](https://en.wikipedia.org/wiki/Edge_computing), lightweight, and low bandwidth communications protocol primarily used by [IoT devices](https://en.wikipedia.org/wiki/Internet_of_things). It was specifically designed for industry, so that's the context in which we'll lay out our brief explanation.
```
chrivers, line 7: Que?

Queue? ;-)

chrivers, line 7: I'm about 0% convinced that Anker's use of MQTT constitutes "Edge driven"/"Edge computing". That is a buzzword-heavy term, with little technical backing, and I'd prefer if we dropped it from this guide.

Same for "stateful". Depending on how you look at it, MQTT is *not* stateful. Rather it's an event transport mechanism. Any number of publishers and consumers can each update and be updated about subscribed topics.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The two actors in this type of network are a **Broker** and a **Client**.  MQTT uses a publish/subscribe model to pass data and commands between clients facilitated by a broker. In order to organize where data and commands are supposed to go, the Broker organizes a hierarchy of **Topics**. 
+
+A Topic is simply a namespace that clients can subscribe and/or publish to. It can describe almost any level of abstraction in your network hierarchy, but always ends at a single client device parameter at the lowest level.
```
chrivers, line 11: "A Topic is simply a namespace that clients can subscribe and/or publish to."

That part is good. The last part of this paragraph is confusing.

Maybe we can compare it to participants in a chat system, such as IRC or discord? That should be a pretty close analogy.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The Broker responsible for receiving client-initiated connections and forwarding messages sent by the client to some other eligible clients.
```
chrivers, line 13: The Broker "is"

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The Clients have the ability to subscribe to Topics in the Broker's directory to pull down any published updates (data or commands) from that point forward. They also have the ability to publish updates to any Topic the Broker gives them access to. 
+
+Clients are responsible for creating any instances of a connection to the Broker and thus the network is "edge-driven". Connections work over TCP and need only one outgoing port per connection instance for 2-way communications.
```
chrivers, line 17: "Clients are responsible for creating any instances of a connection to the Broker and thus the network is "edge-driven". Connections work over TCP and need only one outgoing port per connection instance for 2-way communications."

This is not a good explanation, and I don't think it helps explain what's going on. Let's remove.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The messages consists of only two core parts: the Topic and the **Payload**. While the topic is self-explanatory, the Payload isn't really formally defined by MQTT. It can be anything from a single raw bit to a JSON file.
```
chrivers, line 19: It is formally defined - it just has no structure beyond a series of bytes.

The important part is that it's entire application-specific, and while json is a common choice, some applications chose *ahem* more esoteric formats... We know which ones ;-)

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+Here's an example of a simple MQTT connection that consists of two Clients (Client Ae and Client B), a Broker, and a single Topic of "network/facility/clientA/temp". In this example, this is a simple remote monitoring scenario for a temperature sensor where Client A acts as a publisher and Client be acts as a subscriber.
```
chrivers, line 21: `s/Ae/A/`

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+```mermaid
+graph
+    A("[Broker] Topic: network/facility/clientA/temp") <-- "Publish: network/facility/clientA/temp" --> B(("Client A"))
+    A <-- "Subcribe: network/facility/clientA/temp" --> C(("Client B"))
+
+```
+
+In a more complex network, there is likely many Topics per Client and the Broker will be in charge of handling thousands of topics at once. Clients can also send commands to other clients on the network via Topics to trigger a real world action, such as a heater turning on or a status light changing color.
```
chrivers, line 30: `s/is/are/`

chrivers: also, commas.

chrivers: I think we should remove this paragraph. It doesn't really help the explanation.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+## M5 Specifics
```
chrivers, line 32: Excellent headline - good to see the M5-specific parts clearly delineated :+1:

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+```
+Non-authoritative answer:
+Name:    ankermake-emqx-nlb-60dd062d3b52769a.elb.us-east-2.amazonaws.com
+Addresses:  3.22.247.85
+          3.134.119.140
+          18.189.81.10
+Aliases:  make-mqtt.ankermake.com
+```
+
+As you can see in the `Name:` field, it's an EMQX instance running on an AWS server.
```
chrivers, line 53: Minor nitpick: The *hostname suggests* this, but we can't be sure.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Update Behavior
+
+As long as a M5 is powered on, it publishes updates to the broker servers under the `/phone/maker/YOUR_SERIAL_NUMBER_HERE/notice` topic as a client device. The update seems to be done by exception, meaning that it only updates when a value changes.
```
chrivers, line 57: "The update seems to be done by exception"

That's an odd phrase.

"it only updates when a value changes" <-- good, helpful

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The M5 also subscribes to  commands that may be sent to it from the MQTT servers for execution.
```
chrivers, line 59: I think we can explain this better.

The M5 also subscribes to other topics, where it will react to events, and treat them as commands to perform various actions. In this way, the M5 can be partially remote-controlled over MQTT.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+Another client device (your phone or computer) subscribes to topics associated with the target printer and can publish commands for execution on the target printer.
+
+The basic relationships for MQTT communications can be illustrated as such:
+
+```mermaid
+graph
```
chrivers, line 66: Excellent diagram - really helps figuring out what's going on.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+    A("[Broker] Ankermake MQTT Servers (Job Queue and Data)") <-- "MQTT Data (TLS)" --> B(("M5 Printers"))
+    A <-- "MQTT Data (TLS)" --> C(("Client Devices"))
+    D("AnkerMake HTTPS Servers (API)") -- "MQTT Login Info" --> C
+    C -- "Authentication Token" --> D
+```
+
+We need to utilize the AnkerMake HTTPS API to pull down the MQTT login info and MQTT encryption key needed to setup a new client device connection instance. Your **Authentication Token** (auth_token) is what you need to authenticate your identity with the AnkerMake HTTPS servers. Then, we can plug in those MQTT credentials into the MQTT server to subscribe/publish to the topics in the broker.
```
chrivers, line 73: This is the first time "We" need anything - let's keep the passive voice instead.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The messages are exchanged via an encrypted tunnel (standard MQTT over TLS) but part of the payload is further encrypted using standard AES-256-CBC using a MQTT only encryption key.
```
chrivers, line 75: `s/a MQTT only/an MQTT-only`

I think we can improve the explanation here. We (you and I) both know that the mqtt payloads are encrypted with AES, but this is the reader's first meeting with this concept.

Maybe we should explain that (somewhat unusually) the mqtt packets are encrypted, even though they're sent over a secure channel, that itself requires authentication.

Then we can explain how to get that authentication - i.e. over the http api.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Payload Structure
+
+In the `specifications` folder, you'll find a file named `mqtt.stf` and that's what we're using to map out the contents of the message payload. We don't know a whole lot right now about the possibilities of what you could see in these payloads, but we've managed to map some of the standard scenarios out.
```
chrivers, line 79: Actually, in terms of MQTT package formats, we only have a few magic bytes in the header left to decipher.

Everything else is the application-specific meaning of the json body, after parsing and decryption. And even that, I've gotten a better idea of recently.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+The messages are exchanged via an encrypted tunnel (standard MQTT over TLS) but part of the payload is further encrypted using standard AES-256-CBC using a MQTT only encryption key.
+
+### Payload Structure
+
+In the `specifications` folder, you'll find a file named `mqtt.stf` and that's what we're using to map out the contents of the message payload. We don't know a whole lot right now about the possibilities of what you could see in these payloads, but we've managed to map some of the standard scenarios out. 
+
+The entire payload is delivered as a series of raw bytes and we attempt to unwrap that into something useful. You can do whatever post processing to this packet you like to separate it into useful parts. In our example file, we pretty much just label the entire packet based on our best guess.
```
chrivers, line 81: I don't understand this paragraph.

Are we talking about the general case, where mqtt can contain anything? That's true, but we only have a few unknown bytes left in the message. After parsing and decryption, we always (in every sample I've seen) get perfectly well-formed JSON messages out as a result.

What example file are you talking about?

(It's true that PPPP so far requires a bit more guesswork, but that's unrelated to the mqtt situation)

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+In general, there is two parts to payload structure an unencrypted prefix and an encrypted body. You'll have to separate the prefix from the body, decrypt the body, and then put them back together upon receiving each message. A similar process has to take place on the outbound messages as well.
```
chrivers, line 83: `s/is/are/`

The explanation is correct per se, but it sounds like we don't have code to help the user do all this :)

Maybe we can explain that all this is handled by the protocol implementation in libflagship?

It's also perfectly fine if libflagship is outside the scope of this file. That would probably make some things clearer, but then we should try to describe the protocol in more dry, objective terms.

documentation/example-file-usage/example-file-prerequistes.md
```diff
@@ -0,0 +1,89 @@
+# Prerequisites for using the example scripts
+
+In order to run the example scripts, you'll need to install some supporting software onto your machine (don't worry, it's really easy). 
+
+You'll need Python 3, some pip packages, and the AnkerMake slicer. 
+
+If you're on Linux, you'll also need Wine to install the AnkerMake slicer unless you are ok with ripping files from another machine using MacOS or Windows.
+
+We also include the Transwarp Complier in our `requirements.txt` file for convenience should you want to generate new configs of the standards from this repository.
```
chrivers, line 9: Nice - although we don't generate "configs", we literally generate parts of the python library from the templates.

documentation/example-file-usage/example-file-prerequistes.md
```diff
@@ -0,0 +1,89 @@
+# Prerequisites for using the example scripts
+
+In order to run the example scripts, you'll need to install some supporting software onto your machine (don't worry, it's really easy). 
+
+You'll need Python 3, some pip packages, and the AnkerMake slicer. 
+
+If you're on Linux, you'll also need Wine to install the AnkerMake slicer unless you are ok with ripping files from another machine using MacOS or Windows.
```
chrivers, line 7: Can we mention

1) that these tools are natively supported on Linux
and
2) Wine is an option for running AnkerMake slicer on Linux.

"ripping files" is really up to the user, and outside the scope.

documentation/example-file-usage/mqtt-connect-example-file-usage.md
```diff
@@ -0,0 +1,40 @@
+# Connecting to a M5 over CLI (via MQTT)
+
+This example file connects to your machine over MQTT to view basic data.
+
+1. Navigate to wherever you cloned this repository to. Open the "examples" folder and open a terminal window there, just like in the previous section.
+
+2. Type in the following command, but replace "YOUR_AUTH_TOKEN_HERE" with your actual Authentication Token and no quotes:
+
+   ```bash
+   python3 mqtt-connect.py -r us -A "YOUR_AUTH_TOKEN_HERE"
```
chrivers, line 10: Can we briefly explain `-r eu` vs `-r us`? Just to give people an idea that this is a critically important argument, that they *have* to care about.

chrivers (changes requested):
Big improvements, but still some nits :)

documentation/mqtt-overview.md
austinrdennis: Oops!

documentation/mqtt-overview.md
austinrdennis: Gotcha, I think I was getting confused about how much we know about the messages with PPPP. I'll clean this up

requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+git+https://github.com/chrivers/transwarp.git
```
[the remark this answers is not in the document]
austinrdennis: For what it's worth, we basically did this for me in Windows som I'm confident it works there. I also specifically did this in MacOS and it works, I just have no printer to connect to right now to verify past the script just running to that stage.

documentation/example-file-usage/mqtt-connect-example-file-usage.md
austinrdennis: Ah yes, I totally overlooked that

austinrdennis (commented):
Refined the documentation files according to feedback from @chrivers.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,88 @@
...
+
+The messages are exchanged via an encrypted tunnel (TLS) and that's standard per the MQTT specification. What's unusual about AnkerMake MQTT messages, is that part of the payload is also encrypted before transmission using standard AES-256-CBC. In order for the communications to be successful, the messages must be encrypted and decrypted at either end. The AES encryption of the payload is redundant from a security standpoint (since the entire connection is already encrypted), but it's required for any program interfacing with an M5 to function.
+
+In order to encrypt and decrypt the payload, it's necessary to utilize the AnkerMake HTTPS API. The API allows a client to pull down the MQTT login info and MQTT-only encryption key needed to setup a new client device connection instance and perform cryptography on the payload. 
+
+Your **Authentication Token** (auth_token) is needed to authenticate a client's identity with the AnkerMake HTTPS servers. Then those MQTT credentials are presented to the MQTT server to subscribe/publish to the topics in the broker. Finally the MQTT-only key is used to decrypt incoming messages and encrypt outgoing messages.
```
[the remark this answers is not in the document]
chrivers: As in: the `auth_token` is a secret that the https api hands out, to prevent email+password being stored and used all over.

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,88 @@
...
+graph
+    A("[Broker] Ankermake MQTT Servers (Job Queue and Data)") <-- "MQTT Data (TLS + Encrypted Palyoad)" --> B(("M5 Printers"))
+    A <-- "MQTT Data (TLS + Encrypted Palyoad)" --> C(("Client Devices"))
+    D("AnkerMake HTTPS Servers (API)") -- "MQTT Login Info + Key (HTTPS)" --> C
+    C -- "Authentication Token (HTTPS)" --> D
+```
+
+### Payload Structure
+
+As detailed in the previous section, part of the payload is encrypted. The payload is split into two parts: an unencrypted header and an encrypted JSON body. A MQTT-only key is used to decrypt incoming messages and encrypt outgoing messages per AES-256-CBC. This parsing and cryptography can be accomplished in your programs by using the `libflagship` library in this repository.
```
chrivers, line 86: `encrypt outgoing messages per` -> `encrypt outgoing messages using`

chrivers, line 86:
```
This parsing and cryptography can be accomplished in your programs by using
the `libflagship` library in this repository.
```

...nice :+1: :smile:

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,88 @@
...
+
+In the `specifications` folder, you'll find a file named `mqtt.stf` that maps out the contents of the message payload. There's a few "[magic bytes](https://www.netspi.com/blog/technical/web-application-penetration-testing/magic-bytes-identifying-common-file-formats-at-a-glance/)" in there, but most of the application-specific JSON body is mapped out.
```
chrivers, line 88: Ah, we have some confusing terminology here.

In *many* cases (in libflagship), the purpose of `magic` bytes are entirely known.

The meaning of "Magic" bytes come from convention/decree, rather than value.

For example, PPPP frames starts with `0xF1`. Why `0xF1`? No particular reason, but that's what they chose. So it's "magic" (but known).

Contrast this with other bytes that might encode a length, size, time, loudness, color or similar value type. Those bytes have a fairly predictable meaning, and have very little arbitrary variation.

chrivers (changes requested):
Definitely an improvement :)

Look forward to the next round!

documentation/mqtt-overview.md
austinrdennis: I understand that but my explanation was confusing. I'm just going to remove that part as it's not really necessary to explain here.

spuder:
You may want to rebase. The requirements.txt was merged in #6

documentation/mqtt-overview.md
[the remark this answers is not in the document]
austinrdennis: That's a good point, but I think that belongs in the API overview just to keep it all clean and relavent. It's important to understand for a 100 meter high view of the project, but the scope of this file is more like 20 meters. Am I missing additional uses of the auth-token other than interfacing with the HTTPS API?

documentation/mqtt-overview.md
```diff
@@ -0,0 +1,88 @@
...
+
+A Topic is simply a namespace that clients can subscribe and/or publish to. 
+
+> For example, consider modern chat/messaging applications such as iMessage and Discord. Most of these applications have implemented a feature that shows all the participants of a chat that another participant is typing a response. The topic for this feature may look something like `chat-app/conversation/participant-a/is-typing` and have a boolean value set to "false" by default. 
+>
+> When Participant A starts to type in the compose field, their client would publish an update to this topic that sets the value to "true". Each participant would have a corresponding topic that represents that they are typing and the other participant's clients would be subscribed to these topics. When the value of these topics changes to "true", each participant's client would present an animation to communicate this to the user.
+
+The Broker is responsible for establishing client-initiated connections and forwarding messages sent by the originating client to other subscribed clients.
```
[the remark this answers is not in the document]
austinrdennis: Correct me if I'm misunderstanding, but my reasoning goes like this: For a TLS connection to form, both sides need to handshake. What I'm trying to say here is that the hanshake is intiated by a client and then agreed upon by the broker. The broker never reaches out to a client first, because it has no idea they exist prior to them attempting a connection.

I guess that is kinda obvious and hence why you're asking. How can I re-phrase that to either make it more clear or should I just remove the unecessary info?

documentation/mqtt-overview.md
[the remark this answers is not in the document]
chrivers: You're absolutely right - and I agree that an upcoming "api overview" document would be the right place for this explanation, not here.

documentation/mqtt-overview.md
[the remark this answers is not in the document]
chrivers: Now I see :)

I think we can make it simpler for everybody (including ourselves), by saying something like:

"MQTT messsages are sent over a transport connection between the client and the broker. This connection can be unencrypted or use TLS encryption, and will always be initiated from the client side."

austinrdennis:
> You may want to rebase. The requirements.txt was merged in #6

@spuder Are you saying to rebase from the upstream branch and then merge to master? This is my first collaboration project on GitHub so I don't really know how the branches are being structured at the moment.

austinrdennis:
> You may want to rebase. The requirements.txt was merged in #6

That's on a branch ahead of master, I suggest merge to master before I rebase and merge to master.

spuder:
Depends on your comfort level with git. 

You could change the PR destination branch from master to `chrivers/ankerctl` and then try 
```
 git checkout chrivers/ankerctl
 git pull
 git checkout mqtt-example-documentation
 git merge chrivers/ankerctl
 git push
 ```
 
 Or just delete the requirements.txt

File: documentation/example-file-usage/extract-auth-token-example-file-usage.md [hunk 36]
```diff
@@ -0,0 +1,25 @@
...
+
+1. Navigate to wherever you cloned this repository to and into the `examples` folder in your terminal.
+
+2. MacOS and Windows users (who used the default AnkerMake slicer installation path), type in the following command:
+
+   ```bash
+    python3 extract-auth-token.py -a
+   ```
+
+   For linux users (and MacOS/Windows users that did not use the default installation path), it's a little more complicated since we can't assume where your `login.json` file is. Locate your `login.json` file in your system and use the following command, substituting the path to your file:
```
spuder, line 13: This is no longer true. It now automatically finds the login.json on windows and mac in pr #10 and #7 

No longer need to specify the file

.gitignore
```diff
@@ -1,3 +1,4 @@
 *~
 __pycache__
 .DS_Store
+settings.json
```
chrivers, line 4: Hm, where did this change come from? What is it for?

documentation/example-file-usage/extract-auth-token-example-file-usage.md
```diff
@@ -0,0 +1,25 @@
+# Extracting your Authentication Token
+
+**NOTE:** You need to have to have logged into the AnkerMake slicer in order for this to work, but it doesn't need to be open during this.
```
chrivers, line 3: Excellent - makes sense to mention here.

documentation/example-file-usage/extract-auth-token-example-file-usage.md
```diff
@@ -0,0 +1,25 @@
...
+
+1. Navigate to wherever you cloned this repository to and into the `examples` folder in your terminal.
+
+2. MacOS and Windows users (who used the default AnkerMake slicer installation path), type in the following command:
+
+   ```bash
+    python3 extract-auth-token.py -a
+   ```
+
+   For linux users, it's a little more complicated since we can't assume where your `login.json` file is. Locate your `login.json` file in your system and use the following command, substituting the path to your file:
```
chrivers, line 13: If we're saying "MacOS" and "Windows", we should say "Linux". Or lowercase all the way - as long as it's consistent :)

.gitignore
austinrdennis: This is from my VSCode placing a settings.json in the working directory automatically. It also automatically added this to the gitignore file. I can remove the file and reference to it in the gitignore file if necessary.

chrivers: That's fine, I was just wondering.

But in that case, we should write it as `/settings.json`, otherwise *any* file in this repo named `settings.json` will be ignored, and that might be an actual footgun in the future :)

austinrdennis: Oh, ya I had no idea about that, I'll rebase and commit that change,

PR #6, opened 2023-03-22
Add requirements.txt and fix pppp.py 

spuder:
Fixes error:

```
./ankerctl.py
Traceback (most recent call last):
  File "/Users/<redacted>/Code/github/spuder/ankermake-m5-protocol/./ankerctl.py", line 23, in <module>
    from libflagship.pppp import AnkerPPPPApi
ImportError: cannot import name 'AnkerPPPPApi' from 'libflagship.pppp' (/Users/<redacted>/Code/github/spuder/ankermake-m5-protocol/libflagship/pppp.py)
```

Requirements.txt was taken from this branch https://github.com/Ankermgmt/ankermake-m5-protocol/pull/5/files and then added dependencies `click` and `platformdirs`

PR #7, opened 2023-03-22
Fix Autodiscovery of login.json

spuder:
Fixes autodiscovery of login.json

Depends on #6

### DIFF
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -3,6 +3,8 @@
 import json
 import click
 import logging
+import platform
+from os import path
 from rich import print
 
 import cli.config
@@ -128,7 +130,7 @@ def config_import(env, fd):
     print(cache)
 
 @config.command("import")
-@click.argument("fd", required=True, type=click.File("r"), metavar="path/to/login.json")
+@click.argument("fd", required=False, type=click.File("r"), metavar="path/to/login.json")
 @pass_env
 def config_import(env, fd):
     """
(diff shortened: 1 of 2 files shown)


ankerctl.py
```diff
@@ -138,6 +140,20 @@ def config_import(env, fd):
...
+
+        darfileloc = path.expanduser('~/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json')
+        winfileloc = path.expandvars(r'%LOCALAPPDATA%\Ankermake\AnkerMake_64bit_fp\login.json')
+
+        if useros == 'Darwin' and path.exists(darfileloc):
+            fd = open(darfileloc, 'r')
+        elif useros == 'Windows' and path.exists(winfileloc):
+            fd = open(winfileloc, 'r')
+        else:
+            exit("This platform does not support autodetection. Please specify file location with -f <filename>")
```
chrivers: Good stuff - I was working on a different solution, but I'd like to merge this for now :)

However, we don't use `-f` here. It would be `config import <filename>` instead

chrivers (changes requested):

spuder:
Credit where credit is due. Most of the logic was copied from here: https://github.com/Ankermgmt/ankermake-m5-protocol/blame/master/examples/extract-auth-token.py#L48-L62

ankerctl.py
spuder: I believe this was fixed in the latest commit.

PR #13, opened 2023-03-23
Json output

### LINKED ISSUE #12: json is not machine parseable
jq doesn't like the single quotes that ` ./ankerctl.py config decode` returns

```
./ankerctl.py config decode 2>/dev/null | jq .     
parse error: Invalid numeric literal at line 2, column 14
```

```
jq empty /tmp/foo.json 
parse error: Invalid numeric literal at line 2, column 14
```

## Workaround

- Replace single quotes with double quotes
- Replace booleans like `None` with strings

```bash
./ankerctl.py config decode 2>/dev/null |  sed "s/'/\"/g" | sed 's/None/"None"/g' | jq -r
```

## Solution

Change the single quotes to double quotes

```
{
    'ab_code': 'US',
    'auth_token': 'REDACTED',
    'avatar': '',
```

```
{
    "ab_code": "US",
    "auth_token": "REDACTED",
    "avatar": '',
```  

## Priority

This is a low priority, most people won't need to interact with this file much.

spuder:
Fixes #12 

```
 ./ankerctl.py config decode -v | jq
[*] Loading file..
{
  "ab_code": "US",
 }
```

### DIFF
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -158,7 +158,7 @@ def config_import(env, fd):
     log.info("Loading file..")
 
     cache = libflagship.logincache.load(fd.read())["data"]
-    print(cache)
+    print(json.dumps(cache, indent=4))
 
 @config.command("import")
 @click.argument("fd", required=False, type=click.File("r"), metavar="path/to/login.json")


chrivers:
I actually found #12 to be a bit amusing.

The reason the "json" is not machine parseable.. is because it's python :sweat_smile:

Still, this change makes a lot of sense. It's very nice to be able to chain this into `jq` :)

Merging, and thanks! :+1:

PR #16, opened 2023-03-25
chrivers/protocol support

chrivers:
- gcode command line
 - support for sending semi-raw mqtt commands
 - does proof-of-concept local network detection of printers via pppp
 - mqtt send will try to show the printer response (doesn't always work, since some messages reply with another type, and some don't reply at all)
 - mqtt send with no arguments will now show a list of all known mqtt message types, including numeric values
 - mqtt send now accepts enum names in addition to numeric types
 - added utility code to simplify common operations
 - cleaned up pppp lan-search (thanks @spuder for feedback)


chrivers:
Looks good to me, the author.

Easy-peasy.

PR #19, opened 2023-03-27
Add Dockerfile

spuder:
Add a Dockerfile

### Reasons 

- User may not have python installed
- Possible for future pipelines that run ankerctl as a service/batch job
- Because we can 🤷 

Understandably, this may not be for everyone, or even a direction you want the project to go. But It is here if you want it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,13 +17,14 @@ FROM python:3.11-slim
 # Set the working directory to /app
 WORKDIR /app
 
+RUN mkdir -p /root/.config/
+
 # Copy the script and libraries
-COPY ankerctl.py /app/ankerctl
+COPY ankerctl.py /app/ankerctl.py
 COPY libflagship /app/libflagship
+COPY cli /app/cli
 
 # Copy the installed dependencies from the build environment
 COPY --from=build-env /usr/local/lib/python3.11/site-packages /usr/local/lib/python3.11/site-packages
 
-# Run the script
-# CMD ["python", "ankerctl.py"]
-ENTRYPOINT ["python", "/app/ankerctl"]
+ENTRYPOINT ["python", "/app/ankerctl.py"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+# First stage: build environment
+FROM python:3.11-bullseye AS build-env
+
+# Set the working directory to /app
+WORKDIR /app
+
+# Copy the requirements file
+COPY requirements.txt .
+
+# Install the dependencies
+RUN pip install --no-cache-dir --upgrade pip && \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -141,6 +141,40 @@ Some examples:
 ./ankerctl.py mqtt rename-printer BoatyMcBoatFace
 ```
 
+
+## Docker
+
+While running the python script is generally prefered, there may be situations where you want a more portable solution. For this, a docker image is provided.
+
+```sh
+docker build -t ankerctl . --no-cache
```
chrivers, line 150: Why do we use `--no-cache` here?

What does it do, exactly?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Set the working directory to /app
+WORKDIR /app
+
+RUN mkdir -p /root/.config/
+
+# Copy the script and libraries
+COPY ankerctl.py /app/ankerctl.py
+COPY libflagship /app/libflagship
+COPY cli /app/cli
```
chrivers, line 25: Would this work?

`COPY ankerctl.py libflagship/ cli/ /app/`?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Copy the installed dependencies from the build environment
+COPY --from=build-env /usr/local/lib/python3.11/site-packages /usr/local/lib/python3.11/site-packages
+
+ENTRYPOINT ["python", "/app/ankerctl.py"]
```
chrivers, line 30: Really nice dockerfile - clean separation between build-env and build product.

I'm not very familiar with packaging python in docker images, but would a venv maybe make sense? I suppose dealing with `/usr/local/lib/python3.11/site-packages` should pretty much always work, since we're explicitly depending on `python:3.11...`

README.md
chrivers: Ah, that's actually reasonable. And thanks for the explanation :)

But yes, I think we should get rid of it for the readme, to make the command a little bit simpler for beginners.

Dockerfile
spuder: Yes, tested and changed

chrivers (approved):

Dockerfile
spuder: I'm not really a python expert and I haven't used venv almost at all. 

I checked the python3-slim container and I'm not seeing any virtual environment tools included. I suspect that docker is intending to be the virtual environment. 

```
docker run --entrypoint /bin/bash -it python:3.11-slim
root@3f8b068f8d04:/# venv
bash: venv: command not found
```

chrivers:
Looks good - thank you!

PR #21, opened 2023-03-28
Add a .gitattributes file

spuder:
Add a .gitattributes file to gently coerce windows users to play nice with linux users. 

This shouldn't be a problem if windows dev have set their git config to use core.autocrlf, however in the case when a dev _hasn't_ set that git variable, this will define how to handle line endings correctly. 

Additional information

https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
```
chrivers, line 2: Last line doesn't end in newline :)

chrivers: I just looked at the repo, and we only use `*.(py|tpl|stf|txt|md)` - if we set those to `text`, that should do it?

chrivers: git might actually understand that syntax, but we should probably make 5 lines, each with a simple glob :)

spuder: Ok added specific lines

chrivers (approved):

chrivers:
Merged, thank you!

PR #22, opened 2023-03-28
Force newlines for editorconfig aware editors

spuder:
Some editors (such as vscode) will honor a `.editorconfig` file in a git repo

This tells editorconfig aware editors to force a newline at the end of files

https://editorconfig.org/


.editorconfig
```diff
@@ -0,0 +1 @@
+insert_final_newline = true
```
chrivers, line 1: The line in the editor config, which is about inserting final newlines, it itself missing a final newline.

...the irony, sir, is *staggering*

spuder: 🤔 Can this be done? Did I do it wrong? 🤔

spuder: Looks like users need to have the editorconfig plugin enabled https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

This makes this PR, only useful in certain situations and won't help everyone

chrivers: Looks like most people will need a plugin - but also gitlab and github might pick this up on their own.. so.. Yeah, let's go :)

spuder: Also added a .vscode/settings.json file for the vscode users who _dont_ have the editor config plugin enabled 
This should cover most of the vanilla setups. 

```
{
    "files.insertFinalNewline": true
}

```

chrivers: LGTM, I'll rebase and merge

chrivers (approved):

PR #27, opened 2023-04-03
chrivers/pppp-improvements

chrivers:
Major milestone!

This PR contains a number of new features, mostly focusing on improving the PPPP implementation. PPPP is the proprietary peer-to-peer network that is used by Ankermake printers to upload files, stream video, etc.

Highlights of this PR:
 - Supports automatically upgrading `ankerctl` config files on changes
 - Supports sending print jobs to printer, without any Anker software(!)
 - Supports capturing video stream from camera to local file! (***not yet on newest printer firmware***)

### DIFF
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -231,8 +231,8 @@ def pppp_print_file(env, file, no_act):
             log.error(f"Could not send print job: {E}")
     else:
         log.info("Successfully sent print job")
-
-    api.stop()
+    finally:
+        api.stop()
 
 
 @main.group("http", help="Low-level http api access")
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -426,6 +426,7 @@ def config_show(env):
             print(f"    sn:        {p.sn}")
             print(f"    ip:        {p.ip_addr}")
             print(f"    wifi_mac:  {cli.util.pretty_mac(p.wifi_mac)}")
+            print(f"    api_hosts: {', '.join(p.api_hosts)}")
             print(f"    p2p_hosts: {', '.join(p.p2p_hosts)}")
 
 
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -196,7 +196,8 @@ def pppp(): pass
 def pppp_lan_search(env):
     api = AnkerPPPPApi.open_broadcast(timeout=1.0)
     try:
-        resp = api.req(PktLanSearch())
+        api.send(PktLanSearch())
+        resp = api.recv()
     except TimeoutError:
         log.error("No printers responded within timeout. Are you connected to the same network as the printer?")
     else:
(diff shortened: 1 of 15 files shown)


ankerctl.py
```diff
@@ -165,24 +194,98 @@ def mqtt_gcode(env):
...
     except TimeoutError:
         log.error("No printers responded within timeout. Are you connected to the same network as the printer?")
     else:
         if isinstance(resp, libflagship.pppp.PktPunchPkt):
             log.info(f"Printer [{str(resp.duid)}] is online")
 
+
+@pppp.command("print-file")
+@click.argument("file", required=True, type=click.File("rb"), metavar="<file>")
+@click.option("--no-act", "-n", is_flag=True, help="Upload only (do not start printing)")
```
LazeMSS, line 223: This could be a bit confusing - one might think it uploads and stores the file - so maybe add something like "for testing upload"

chrivers: Good point - I'll fix

spuder: `--no-print` or `--dry-run` may be more intuitive semantics over `--no-act`

ankerctl.py
```diff
@@ -260,84 +367,68 @@ def config_import(env, fd):
...
-            exit("This platform does not support autodetection. Please specify file location")
+        try:
+            if useros == 'Darwin':
+                fd = open(darfileloc, 'r')
+            elif useros == 'Windows':
+                fd = open(winfileloc, 'r')
+            else:
+                log.critical("This platform does not support autodetection. Please specify file location")
+        except FileNotFoundError:
+            log.critical("Failed to import file - check if you are logged into Ankerslicer")
```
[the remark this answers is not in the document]
chrivers: Yeah, the code may be a bit unintuitive here. But I added an extension to the logging system, such that `log.critical()` always aborts after printing the error.

Otherwise we'd have to raise an exception, and then catch it somewhere else (which is a bit tricky when using click)

ankerctl.py
chrivers: How do you feel about the naming in the newest version?

chrivers: How about now?

ankerctl.py
```diff
@@ -260,84 +371,68 @@ def config_import(env, fd):
...
+    log.info("Finished import")
 
 
 @config.command("show")
 @pass_env
 def config_show(env):
     """Show current config"""
 
+    log.info(f"Loading config from {env.config.config_path('default')}")
+    print()
```
LazeMSS, line 412: ?

chrivers: For aesthetics :)

When printing the config, the rest of the info is clearly divided into two sections, so this looks nicer, imho.

spuder (approved):

PR #29, opened 2023-04-04
Add ability to change host/port

spuder:
## Changes

- Adds the ability to set --host and --port when starting the flask server
- Fix path issue with COPY on docker build
- Adds an example to run webserver in docker

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,28 @@
+version: '3.9'
+
+# You must manually run `ankerctl config import` atleast once
+# docker run \
+# -v ankerctl_vol:/root/.config/ankerctl \
+# -v "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json" \
+# ankerctl config import /tmp/login.json
+services:
+  ankerctl:
+    image: ankerctl/ankerctl:latest
+    container_name: ankerctl
+    restart: unless-stopped
+    build: .
+    environment:
+      - FLASK_PORT=4470
+      - FLASK_HOST=0.0.0.0
+    volumes:
+      - ankerctl_vol:/root/.config/ankerctl
+    ports:
+      - 4470:4470
+    # entrypoint: "/bin/sleep"
+    # command: ["infinity"]
+    entrypoint: "/app/ankerctl.py"
+    command: ["webserver", "run"]
+
+volumes:
+  ankerctl_vol:
+    external: true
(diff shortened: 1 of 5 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,28 @@
...
+# ankerctl config import /tmp/login.json
+services:
+  ankerctl:
+    image: ankerctl/ankerctl:latest
+    container_name: ankerctl
+    restart: unless-stopped
+    build: .
+    environment:
+      - FLASK_PORT=4470
+      - FLASK_HOST=0.0.0.0
```
chrivers, line 16: I guess if we use `127.0.0.1` inside docker, that'll be a problem, since we can't access it from the host.

But this might be a little too exposed, since we have no authentication of any kind.

We should probably mention in the README, that running this docker-compose file will expose the API server on 4470, or?

I'm open to suggestions

chrivers (commented):
Looks really good - just one minor question about how much we're exposing the service.

docker-compose.yaml
spuder: Changing the flask bind ip from 127.0.0.1 to 0.0.0.0 is required if running inside a docker container because each docker container gets its own isolated network with its own isolated looback address. 

Here is proof to show that despite _appearing_ to be open to the world, the docker container is still only available at 127.0.0.1. The real magic is this line
```
    ports:
      - 127.0.0.1:4470:4470
```

## Tests

All these tests pass, showing the container is only available at `127.0.0.1:4470` and `localhost:4470`
```
docker compose up
check_curl() {
  if curl --head --silent --fail --output /dev/null --max-time 2 "http://$1:$2"; then
    echo "✅"
  elif [ $? -eq 28 ]; then
    echo "❌ (timeout)"
  else
    echo "❌"
  fi
}



check_curl 127.0.0.1 4470
✅
check_curl 127.0.0.1 5000
❌
check_curl localhost 4470
✅
check_curl localhost 5000
❌
check_curl 172.18.0.2 5000
❌ (timeout)
check_curl 172.18.0.2 4470
❌ (timeout)
check_curl 192.168.1.96 5000
❌
check_curl 192.168.1.96 4470
❌

```




## Additional information

- https://www.freecodecamp.org/news/how-to-dockerize-a-flask-app/
- https://www.youtube.com/watch?v=4uoWRXuYfJs

chrivers: Oh my god, I had a noob momemt... I can't believe I missed that :joy:

That absolutely makes sense - and excellent work! I'll get this merged :)

chrivers (approved):

PR #30, opened 2023-04-04
Improve webserver docker

spuder:
## Changes

- ~Moves nginx.conf out of root directory into the templates directory for cleanliness~
- ~Improves the nginx config to use 'stream' instead of normal http load balancing (future proofing for video stream and any other UDP stuff we may want to do)~
- Adds some documentation on the webserver


docker-compose.yaml
```diff
@@ -11,7 +11,7 @@ services:
     container_name: ankerctl_nginx
     restart: unless-stopped
     volumes:
-      - ./nginx.conf:/etc/nginx/nginx.conf
+      - ./templates/docker/nginx.conf:/etc/nginx/nginx.conf
```
chrivers, line 14: I'm all for moving things out of the root dir, but `templates/` is definitely not the way to go. That dir is exclusively for the `transwarp` templates, used to compile the files in `specification/` into rendered code files (python, for now).

How about we make a top-level `docker/` dir, and put everything docker-related there?

templates/docker/nginx.conf
```diff
@@ -0,0 +1,16 @@
...
+}
+
+stream {
+  upstream ankerctl {
+    server ankerctl:5000;
+  }
+
+  server {
+    listen 4470;
+    proxy_pass ankerctl;
```
chrivers, line 14: At this point, we are just letting nginx forward tcp connections. Why is it that we are even using nginx?

In the future, we probably want to run the flask app using uwsgi, and then nginx will be highly useful. But right now I'm not sure I'm seeing the appeal?

chrivers (commented):
Looks good! How do you feel about moving docker-related files do `docker/` ?

chrivers:
That's fair - definitely one of those "sounds easy, but actually hard".

Can you force push this branch without the last commit? Then we can merge.

chrivers (approved):

PR #31, opened 2023-04-04
Support login.json in root app folder on windows

LazeMSS:
Basically allow both the root folder and 64bit folder support when importing the config.json on windows


spuder:
Thanks, Line 345 also needs to be updated (both `config import` and `config decode` use the same logic)

LazeMSS:
> Thanks, Line 345 also needs to be updated (both `config import` and `config decode` use the same logic)
Fixed and also added the same "logic" for the error messages.

spuder (approved):

chrivers:
LGTM - thanks for your contribution :)

I'll see if I can rework that shared logic, so we can avoid the duplication. But this is good for now :+1:

PR #41, opened 2023-04-08
Add pipeline to publish docker container and release

Cisien:
on tag, this pipeline will build a docker container for common architectures (amd64, 32-bit arm, 64-bit arm) and zip up the repo as a .zip file and a .tar.gz file

the zip step is arguably not any different from the default source code zip files that are attached to a release by default, but may be useful placeholders for any kind of build steps that produce os specific packages.

How to use:
As you push changes, or new features, update CHANGELOG.md for the version that's under development, or under the Unreleased header. At some point before creating and pushing a new tag, push this changelog with the next version number as a section with all of the changes for that release. The contents of this header become the release notes.

edit CHANGELOG.md
`git add CHANGELOG.md`
`git commit -m "whatevs" `
`git push`

`git tag v1.2.3`
`git push origin v1.2.3`

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -1,0 +1,92 @@
+name: Docker Image CI
+
+on:
+  push:
+    tags:
+     - v**
+  pull_request:
+    branches: [ master ]
+     
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
+      
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+      
+    - name: Log into registry ${{ env.REGISTRY }}
+      if: github.event_name == 'push'
+      uses: docker/login-action@v2.1.0
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Docker Setup QEMU
+      uses: docker/setup-qemu-action@v2.1.0
+      
+    - name: Docker Setup Buildx
+      uses: docker/setup-buildx-action@v2.5.0
+      with:
+        version: latest
+        install: true
+        
+    - name: Docker Metadata action
+      id: meta
+      if: github.event_name == 'push'
+      uses: docker/metadata-action@v4.3.0
+      with:
+        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+        flavor: |
+          latest=${{ github.event_name == 'push' }}
+        tags: ${{ github.ref_name }}
+        
+    - name: Build and push Docker images
+      uses: docker/build-push-action@v4.0.0
+      with:
+        context: .
+        file: Dockerfile
+        platforms: linux/arm/v7,linux/arm64,linux/amd64
+        push: ${{ github.event_name == 'push' }}
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
+    
+    - name: Create release directory
+      run: mkdir release
+      
+    - name: Zip Release
+      if: github.event_name == 'push'
+      uses: TheDoctor0/zip-release@0.7.1
+      with:
+        type: 'zip'
+        filename: release/ankerctl.zip
+        exclusions: '*.git* release'
+        
+    - name: GZip Release
+      if: github.event_name == 'push'
+      uses: TheDoctor0/zip-release@0.7.1
+      with:
+        type: 'tar'
+        filename: release/ankerctl.tar.gz
+        exclusions: '*.git* release'
+        
+    - name: Release
+      if: ${{ github.event_name == 'push' }}
+      uses: docker://antonyurchenko/git-release:v5.0.1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: |
+          release/ankerctl.zip
+          release/ankerctl.tar.gz
+  
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -69,15 +69,15 @@ jobs:
       uses: TheDoctor0/zip-release@0.7.1
       with:
         type: 'zip'
-        filename: release/ankerctl.zip
-        exclusions: '*.git* release'
+        filename: release/ankerctl-${{ github.ref_name }}.zip
+        exclusions: '*.git* /*release/*'
         
     - name: GZip Release
       if: github.event_name == 'push'
       uses: TheDoctor0/zip-release@0.7.1
       with:
         type: 'tar'
-        filename: release/ankerctl.tar.gz
+        filename: release/ankerctl-${{ github.ref_name }}.tar.gz
         exclusions: '*.git* release'
         
     - name: Release
@@ -86,7 +86,5 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
-        args: |
-          release/ankerctl.zip
-          release/ankerctl.tar.gz
+        args: release/ankerctl*
   
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [1.0.0] - 2023-04-07
+


.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+    branches: [ master ]
+     
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
```
chrivers, line 17: I assume `packages` write permission is for publishing releases?

What is `contents` write permission used for?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+  pull_request:
+    branches: [ master ]
+     
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
chrivers, line 16: Building on a moving target could break the build, even if we change nothing.

I'd rather target a smaller, stable image. For example `debian:11`.

What do you think?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
+name: Docker Image CI
+
+on:
+  push:
+    tags:
+     - v**
+  pull_request:
+    branches: [ master ]
```
chrivers, line 8: If I understand this right, we are triggering the build process on (I assume) merged pull requests? Or on all pull requests targeting master?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
+      
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
```
chrivers, line 25: We shouldn't strictly need this, but I guess it doesn't hurt.

Since we got the transwarp dependency listed in `requirements.txt` now, the submodule has become obsolete. I think we'll remove it over time.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+      
+    - name: Log into registry ${{ env.REGISTRY }}
+      if: github.event_name == 'push'
```
chrivers, line 28: Just for my understanding, can you explain what this line does? Why do we need to limit login to `push` actions?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+      uses: docker/login-action@v2.1.0
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Docker Setup QEMU
+      uses: docker/setup-qemu-action@v2.1.0
```
chrivers, line 36: I assume we need qemu for the docker build? Or?

Also, it looks like `setup-qemu-action` takes an optional list of platforms to target. Can we make the build process more light-weight by restricting the platforms to the ones we actually use?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+    
+    - name: Create release directory
+      run: mkdir release
+      
+    - name: Zip Release
+      if: github.event_name == 'push'
+      uses: TheDoctor0/zip-release@0.7.1
+      with:
+        type: 'zip'
+        filename: release/ankerctl.zip
```
chrivers, line 72: I'd strongly prefer we use a descriptive filename. We're going to end up with a hundred files all called `ankerctl.(zip|tar.gz`.

I'm sure there's an environment var that roughly correlates to something like `git describe`?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+        exclusions: '*.git* release'
+        
+    - name: GZip Release
+      if: github.event_name == 'push'
+      uses: TheDoctor0/zip-release@0.7.1
+      with:
+        type: 'tar'
+        filename: release/ankerctl.tar.gz
```
chrivers, line 80: Same here, obviously

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,92 @@
...
+        exclusions: '*.git* release'
+        
+    - name: Release
+      if: ${{ github.event_name == 'push' }}
+      uses: docker://antonyurchenko/git-release:v5.0.1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: |
+          release/ankerctl.zip
```
chrivers, line 90: ..and same here

Unless the build environment is completely cleared between each run, in which case we might be able to just say `release/ankerctl-*`?

chrivers (commented):
Awesome work - thanks for your contribution :)

It'll be really exciting to get an automatic build system going.

Also, I'd be very nice to have a way for the tool itself to know its own version. Do you have any suggestions? (doesn't have to be in this PR)

.github/workflows/build-and-publish.yml
spuder: This is an interesting one. I’m a hard proponent of always pinning versions, yet GitHub actions runs-on stanza might be worth an exception for the following reasons:

1. All actions documentation seems to recommend using either ‘ubuntu-latest’, ‘windows-latest’ or macos-latest’. I’ve never seen an exception to this outside of some nuanced osx builds targeting old versions. Here are the docs: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
2. the runs on stanza isn’t referencing the official ubuntu container like you  might expect. It is actually a blessed image from GitHub. See this quote 

> Note: The -latest runner images are the latest stable images that GitHub provides, and might not be the most recent version of the operating system available from the operating system vendor.

3. Anecdotally I’ve been using actions since 2018 and I’ve never seen a build break from ubuntu-latest.

.github/workflows/build-and-publish.yml
Cisien: any pull request targeting master -- this is a PR verification step mostly. Any actual publishing steps are skipped for these builds.

.github/workflows/build-and-publish.yml
Cisien: likewise, I've been using ubuntu-latest since the dawn of actions and haven't had any issues from it. Here's the repo that covers the runner images.
https://github.com/actions/runner-images

latest is currently ubuntu-22.04. For linux runners, we have a choice between latest (2204), 2204 pinned, and 2004

.github/workflows/build-and-publish.yml
Cisien: publishing releases, a setting on the project needs to be flipped as well.

Under Project settings, Actions, General:
image

.github/workflows/build-and-publish.yml
Cisien: the `event_name` is `push` when a new tag is pushed (or technically any commit, but we're only looking at tags). For builds triggered from a pull request, the `event_name` is `pull_request` iirc.

This lets us skip publishing-related steps for builds that will never publish artifacts.

.github/workflows/build-and-publish.yml
Cisien: The qemu step is used by buildx to build docker images for other platforms. We are targeting amd64, 32 bit arm, and 64 bit arm in the docker build step. I don't think targeting platforms here will save us much since this task only takes a few seconds to run.

.github/workflows/build-and-publish.yml
Cisien: you're thinking something like `ankerctl-v1.0.0.(zip|tar.gz)`? I'm pretty sure the tag that triggered the build can be pulled in here.

.github/workflows/build-and-publish.yml
chrivers: That's all reasonable - let's stick with with `ubuntu-latest`. Thanks for the explanation!

.github/workflows/build-and-publish.yml
chrivers: Sounds excellent - it would also allow us to include some testing steps at a later point, right?

.github/workflows/build-and-publish.yml
chrivers: That all makes sense - thanks.

.github/workflows/build-and-publish.yml
chrivers: That makes sense - thanks :)

.github/workflows/build-and-publish.yml
chrivers: That would be great, yeah :)

I'm marking this as "unresolved", since it's not yet implemented.

.github/workflows/build-and-publish.yml
chrivers: So `packages` is for uploading the artefact, but `contents` is for creating a release object? It kind of looks like `contents` also includes complete write access to the whole repository, but I guess we can live with that?

chrivers (changes requested):
Lots of stuff well explained - thanks :)

I'd like to see the naming fixed to include the tag (or other descriptive identifier) before we merge this, but otherwise I feel like we are almost there.

.github/workflows/build-and-publish.yml
Cisien: Yeah, not ideal. Fortunately, github doesn't allow any ol' random PR access to the GITHUB_TOKEN by default.

.github/workflows/build-and-publish.yml
Cisien: I'm going to leave it for now, we an nuke it when the submodule goes away, or just let it stick around in case we need submodules in the future. As you said, it doesn't hurt anything

chrivers (approved):

PR #46, opened 2023-04-10
Docker improvements

billyjbryant:
### Description

Improves a few areas for running as a Docker container:

1. Adds a `CMD` statement to the Dockerfile so that unattended docker deployments (such as through Portainer [or any 3rd party Docker manager]) don't automatically fail due to an invalid command.
2. Changes `configHost` and `configPort` to `requestHost` and `requestPort` and updates how these values are populated. This now permits displaying the IP address and port of the request in the web interface instead of `0.0.0.0:4470` when using `FLASK_HOST=0.0.0.0`. 

### Testing
I tested these changes by building the docker image using Portainer and deploying it to my Synology NAS

#### Webserver view
image

#### Print success
image

#### Container
image

### DIFF
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -458,7 +458,7 @@ app = Flask(__name__, template_folder='./static')
 
 @app.get("/")
 def app_root():
-    return render_template("index.html", configPort = app.config["port"], configHost = app.config["host"])
+    return render_template("index.html", requestPort = request.host.split(':')[1], requestHost = request.host.split(':')[0])
 
 @app.get("/api/version")
 def app_api_version():
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -458,7 +458,9 @@ app = Flask(__name__, template_folder='./static')
 
 @app.get("/")
 def app_root():
-    return render_template("index.html", requestPort = request.host.split(':')[1], requestHost = request.host.split(':')[0])
+    host = request.host.split(':')
+    requestPort = host[1] if len(host) > 1 else '80'
+    return render_template("index.html", requestPort = requestPort, requestHost = host[0])
 
 @app.get("/api/version")
 def app_api_version():
--- a/ankerctl.py
+++ b/ankerctl.py
@@ -459,7 +459,7 @@ app = Flask(__name__, template_folder='./static')
 @app.get("/")
 def app_root():
     host = request.host.split(':')
-    requestPort = host[1] if len(host) > 1 else '80'
+    requestPort = host[1] if len(host) > 1 else '80' # If there is no 2nd array entry, the request port is 80
     return render_template("index.html", requestPort = requestPort, requestHost = host[0])
 
 @app.get("/api/version")
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -10,7 +10,6 @@ services:
     image: ankerctl/ankerctl:latest
     container_name: ankerctl
     restart: unless-stopped
-    pull_policy: build
     build: .
     environment:
       - FLASK_PORT=4470
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,4 +1,4 @@
-version: '3.9'
+version: '3.9.1'
 
 # You must manually run `ankerctl config import` atleast once
 # docker run \
@@ -10,6 +10,7 @@ services:
     image: ankerctl/ankerctl:latest
     container_name: ankerctl
     restart: unless-stopped
+    pull_policy: build
     build: .
     environment:
       - FLASK_PORT=4470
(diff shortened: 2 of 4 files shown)


docker-compose.yaml
```diff
@@ -10,6 +10,7 @@ services:
     image: ankerctl/ankerctl:latest
     container_name: ankerctl
     restart: unless-stopped
+    pull_policy: build
```
chrivers, line 13: We talked about this in the last PR version, but can you elaborate again on what this does/fixes/improves?

1) Are there implications other than the (relatively niche) case of Portainer building from a git url?

2) What does this mean for people who try to build with `docker-compose` locally? Will it require internet access? Will it try to clone the repository again, if you do this on a fresh clone?

ankerctl.py
```diff
@@ -458,7 +458,7 @@ def webserver(env):
 
 @app.get("/")
 def app_root():
-    return render_template("index.html", configPort = app.config["port"], configHost = app.config["host"])
+    return render_template("index.html", requestPort = request.host.split(':')[1], requestHost = request.host.split(':')[0])
```
chrivers, line 461: The flask documentation states (https://tedboy.github.io/flask/generated/generated/flask.Request.host.html#flask.Request.host)

```
Request.host

    Just the host including the port if available. See also: trusted_hosts.
```

So when running on a standard port, this will make the application crash and become non-functional.

Flask uses werkzeug, with the relevant parts in `werkzeug/src/werkzeug/sansio/request.py` and `werkzeug/src/werkzeug/sansio/utils.py`. Maybe we can get inspiration there.

Can you suggest a more robust solution? I like the idea overall :)

chrivers (commented):
Thanks for returning for round 2 :)

Some comments and questions, but nothing too major. Let me know what you think.

docker-compose.yaml
Cisien: To your second question: It will run the docker build command locally against the cloned source. This is probably ideal today since we're not publishing the container image to docker hub, but if we decide to do that later, then this will be a (mostly) redundant step.

I would prefer if we didn't set a pull policy (specifically if we decide to push this image to docker hub) as this is dependent on the user's setup.

chrivers: Does that mean, that if we push this to say, the github docker registry, and somebody uses this image in a `docker-compose.yml` file, the build process will try to make the image from git, instead of using the image we published?

Cisien: yes
https://docs.docker.com/compose/compose-file/05-services/#pull_policy

spuder: `pull_policy: build` seems to be a somewhat recent addition to docker compose. 

Pros:
- Users running local development will always have the latest code baked into their docker containers

Cons:
- `docker compose up` will be slow because it will rebuild the container every time
- `ankerctl:latest` for me isn't necessarily `ankerctl:latest` for you (effectively bypassing the public image). This _shouldn't_ be a problem for most people, as long as that is understood when debugging/troubleshooting

It sort of comes down to the purpose of the docker compose. If it is to make it easy for end users to get ankerctl up and running, or if it is targeted twards developers who want to test inside a container

Alternatives:
1. Have 2 docker compose files, one for end users, and one for developers
- docker-compose.yaml
- docker-compose-dev.yaml

2. Pass in the build_arg as an optional argument. (Something like this):
```
docker compose up --build-arg PULL_POLICY=build
```
or
```
export COMPOSE_PULL_POLICY=build; docker compose up
```

ankerctl.py
billyjbryant: It doesn't cause the application to crash, just the webpage when sourced at `/` (returns a 500); however, I did find a solution for this and have pushed the necessary change.

docker-compose.yaml
billyjbryant: I can test to see if passing it as an optional `build_arg` works for me in Portainer, if not, then I can build images manually at CLI if I need to (if we see this as too much of a disruptive change). This also depends on whether or not there are ever going to be published container images and if so, how frequent these images will be updated.

chrivers: @billyjbryant Any news on this?

Since @Cisien got the github action PR merged, we are almost ready to publish images.

After listening to the inputs - and thank you all again for the detailed arguments - it's clear that this will not be merged as long as `pull_policy: build` is included.

We are trading a lot of potential issues and confusion, for a small convenience for some users. That's not worth it.

If there's some way to get convenience for everybody, I'm open to suggestions, but otherwise this needs a bit of rework.

billyjbryant:
@chrivers I removed `pull-policy` from the docker compose file.

chrivers (approved):
Looks good now - thanks for your contribution :)

PR #50, opened 2023-04-17
Added printing of the printers ip-address when using lan-search.

kivinen:
When using the lan-search command it would be useful to know he ip-address of the printer. Added printing the ip-address of the packet we received from the printer.


chrivers:
Agreed, that seems very reasonable! Thanks for your contribution :)

chrivers (approved):

PR #52, opened 2023-04-25
Improving Web Interface

billyjbryant:
### Description

Currently, if you start the web interface without having previously imported your login configuration, the application crashes. This PR introduces improvements that allow you to upload your login configuration from the web interface. This improves the user experience for running the application in a Docker container or on a remote host. In addition, there is a complete overhaul of the UI to improve the look and feel of the application. 

#### Tangentially
* Introduced new `web.util` and `web.config` modules to supply utilities and config functions to the web interface. 
* Created `docstrings` for all functions in the `web` module
* Implemented route to allow for uploading AnkerMake login file and reloading services once the configuration is imported
* Introduced `.prettierrc` for controlling prettier when formatting non-python files
* Introduced `setup.cfg` for editors that do not recognize `.pep8` for `pycodestyle` formatting
* Cleaned up formatting in everything I touched


billyjbryant:
Walkthrough

billyjbryant:
CC: @chrivers @Cisien

billyjbryant:
Rebased master into this PR, which was a bad move.

billyjbryant:
Resolved merge conflicts

static/index.html
```diff
@@ -4,7 +4,34 @@
...
+
+             $('#copyFilePath').on('click',function(){
+                 navigator.clipboard.writeText("{{ loginFilePath|replace('\\', '\\\\') }}");
+                 console.log('Copied {{ loginFilePath }} to clipboard...')
+                 return false;
+             });
+         });
+        </script>
+        <style>
+         .nav-link:hover {
```
chrivers, line 29: This seems a bit random. What is it for, and why is it inline?

If we need it, please create `static/ankersrv.css`, and put the declarations there.

web/__init__.py
```diff
@@ -1,23 +1,43 @@
...
+import libflagship.httpapi
+import libflagship.logincache
 
 from libflagship.pppp import P2PSubCmdType, FileTransfer
 from libflagship.ppppapi import FileUploadInfo, PPPPError
 
 
 from web.lib.service import ServiceManager
 
+from user_agents import parse
```
chrivers, line 22: With a name as generic as `parse`, that we use only 1 place, we should `import user_agents`, and call `user_agents.parse()` where needed. That will be much easier to follow.

Also, the rough order of imports should be:

 - python builtins
 - modules from pip dependencies
 - modules from this package

web/__init__.py
```diff
@@ -1,23 +1,43 @@
...
+
 import cli.util
+import cli.config
+
+
+def dict_match(dict, match):
+    for key in dict:
+        if re.match(rf"^{key}.*", match):
```
chrivers, line 30: `re.match` matches a substring from the start of the string, so this line is the same as:

```python
if re.match(key, match): ...
```


Also, we only match against non-regex strings, so this test becomes:

```python
if match.startswith(key): ...
```

At that point, this function becomes very light :)

I can see it's used to determine a platform-specific hint about file locations, for now.

How about we make a function that return a standardized platform name:

```python
def user_agent_platform(...):
   if ...:
     return "linux"
   ...
     return "windows"
   ...
     return "macos"
```

Let's put it in `web/utils.py`. Then, the result from this can simply be used in a dictionary lookup.

web/__init__.py
```diff
@@ -1,23 +1,43 @@
...
+            return key
+
 
 app = Flask(
     __name__,
     root_path=".",
     static_folder="static",
     template_folder="static"
 )
+app.secret_key = token(24)
```
chrivers, line 40: What's this used for? Isn't this for cookie encryption (at least among other things?)

web/__init__.py
```diff
@@ -32,60 +52,77 @@
 
 @app.before_first_request
 def startup():
-    app.svc.register("pppp", web.service.pppp.PPPPService())
-    app.svc.register("videoqueue", web.service.video.VideoQueue())
-    app.svc.register("mqttqueue", web.service.mqtt.MqttQueue())
-    app.svc.register("filetransfer", web.service.filetransfer.FileTransferService())
+    if app.config["login"]:
```
chrivers, line 55: I don't think this is a good way to do it.

These services are started on-demand, and should be registered as soon as possible, no matter if we are logged in or not.

Or course, we should be careful about not *requesting* them until we have login information ready.

web/__init__.py
```diff
@@ -32,60 +52,77 @@
...
+        for msg in app.svc.stream("videoqueue"):
+            sock.send(msg.data)
 
 
 @sock.route("/ws/ctrl")
 def ctrl(sock):
 
-    while True:
-        msg = json.loads(sock.receive())
+    if app.config["login"]:
```
chrivers, line 82: Instead of indenting everything in all of these websocket functions, let's follow this pattern instead:

```python
if not app.config["login"]:
    return
```

That way, the core logic isn't pushed needlessly to the right.

web/__init__.py
```diff
@@ -32,60 +52,77 @@
...
 @app.get("/")
 def app_root():
+    config = app.config["config"]
+    user_agent = parse(request.headers.get('User-Agent'))
+    login_path = {
+        'Mac OS': '~/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json',
+        'Windows': r'%LOCALAPPDATA%\Ankermake\AnkerMake_64bit_fp\login.json',
+        'None': 'Unsupported OS, supply path to login.json',
+    }
+    useros = dict_match(login_path, user_agent.os.family)
```
chrivers, line 115: Once we have the `user_agent_platform` function, we can move this logic to `web/platform.py` as well, and get a function that returns a file path hint from a platform name.

something like
```python
platform_name = platform.user_agent_platform(request.headers.get('User-Agent'))
login_path = platform.login_path(platform_name)
```

web/__init__.py
```diff
@@ -32,60 +52,77 @@
...
+
     host = request.host.split(':')
     requestPort = host[1] if len(host) > 1 else '80' # If there is no 2nd array entry, the request port is 80
     return render_template(
         "index.html",
         requestPort=requestPort,
-        requestHost=host[0]
+        requestHost=host[0],
+        configure=app.config["login"],
+        loginFilePath=login_path[useros] if useros in login_path else login_path["None"],
```
chrivers, line 124: I know you're only following the naming scheme here, but it is unfortunately inconsistent with everything else.

Please change to `snake_case` instead of `smallCamelCase`.

web/__init__.py
```diff
@@ -32,60 +52,77 @@
...
+        ankerConfig=str(config_show()) if app.config["login"] else '<p>No printers found, please load your login config...</p>',
```
chrivers, line 125: We shouldn't do server-side rendering unless the alternative is worse.

Let's move this to the template? We should just pass the data structure, then it can do the conditionals there.

web/__init__.py
chrivers: Also, what is `login_path["None"]`? Is it an error? Seems confusing.

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
     }
 
 
+def allowed_file(filename, ALLOWED_EXTENSIONS):
```
chrivers, line 138: a parameter is not a constant. No `SCREAMING_SNAKE_CASE` here.

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+    auth_token = cache["auth_token"]
+
+    # extract account region
+    region = libflagship.logincache.guess_region(cache["ab_code"])
+
+    try:
+        newConfig = cli.config.load_config_from_api(auth_token, region, False)
+    except libflagship.httpapi.APIError as E:
+        flash(f"Config import failed: {E} (auth token might be expired: make sure Ankermake Slicer can connect, then try again", 'danger')
+        return redirect('/')
```
chrivers, line 179: Having to do `return redirect('/')` in all exit paths is a sign something is a little off.

Split this function in 2: one that does the core logic, and one that calls the first one, catches exceptions, and does `flash()`, `redirect()`, etc.

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+    try:
+        config.save("default", newConfig)
+    except Exception as E:
+        flash(f"Config import failed: {E}", 'danger')
+        return redirect('/')
+    flash("Configuration imported! Restarting...", 'success')
+    return redirect('/')
+
+
+@app.post('/api/config/upload')
```
chrivers, line 193: The other `/api` endpoints emulate octoprint, but this is ankerctl-specific.

Move the endpoint to `/api/ankerctl/...` so we don't risk any naming conflicts.

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+def app_api_config_upload():
+    ALLOWED_EXTENSIONS = set(['json'])
```
chrivers, line 195:
```python
ALLOWED_EXTENSIONS = {"json"}
```

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+    with tempfile.TemporaryDirectory(prefix='ankerctl_') as tmpdir:
+        config = app.config["config"]
+
+        if request.method == 'POST':
```
chrivers, line 199: We get here by hooking `@app.post`, so this is always true?

If we want to keep it (for clarity, or otherwise), negate the check and make it return.

```python
if request.method != "POST":
    return
```

(again, to avoid long indentation)

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+            if 'loginFile' not in request.files:
```
chrivers, line 200: Same pattern as above. Split out into a function with pure business logic, and one that handles `flash()`, `redirect()`, etc.

Also, `login_file` while you're at it :)

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+                flash('No file found', 'danger')
+                return redirect('/')
+            file = request.files['loginFile']
+            if file.filename == '':
+                flash('No file selected', 'danger')
+                return redirect('/')
+            if file and allowed_file(file.filename, ALLOWED_EXTENSIONS):
```
chrivers, line 207: Why are we trying to limit what the file can be called?

We should only care if we can read and parse the file. If the user wants to upload `login.bak`, let's not stand in the way.

web/__init__.py
```diff
@@ -98,6 +135,86 @@ def app_api_version():
...
+                filename = secure_filename(file.filename)
+                filepath = os.path.join(tmpdir, filename)
+                file.save(filepath)
```
chrivers, line 210: This `file` reference already works (almost) like a regular file. If we change `config_import()` to accept a file-like object, we can avoid copying it to a temp dir altogether.

web/__init__.py
```diff
@@ -115,9 +232,24 @@ def app_api_files_local():
...
+@app.get("/reload")
+def reload_webserver():
+    config = app.config["config"]
+    with config.open() as cfg:
+        app.config["config"] = config
+        app.config["login"] = True if cfg else False
+        if cfg:
+            session['_flashes'].clear()
+            flash('Configuration loaded', 'success')
+            startup()
```
chrivers, line 244: This is bad - in the current code, this would register all services twice, which will cause an exception.

Register the services always, and just load the config.

I'm not convinced this function/endpoint is needed at all.

ankerctl.py
```diff
@@ -451,15 +454,14 @@ def config_show(env):
 @main.group("webserver", help="Built-in webserver support")
 @pass_env
 def webserver(env):
-    env.require_config()
+    env.require_config(False)
```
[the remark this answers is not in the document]
billyjbryant: we `require_config` by default, unless you pass it the value of `False` :D I think it makes sense. but we could change it to that too :D

static/ankersrv.js
```diff
@@ -56,9 +56,13 @@ $(function () {
         return false;
     });
 
+    let alert_list = document.querySelectorAll('.alert');
+    alert_list.forEach(function(alert) {
+        new bootstrap.Alert(alert);
 
-    $('#configData').on('click',function(){
```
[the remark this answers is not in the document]
billyjbryant: Because inside the Js the variables weren't actually evaluating. Pretty sure files that aren't directly passed through flask don't evaluate

ankerctl.py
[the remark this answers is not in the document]
chrivers: It's just weird to "require" something that isn't "required" :D

Let's rename to `load_config`, then `required={True|False}` makes sense

static/index.html
```diff
@@ -4,7 +4,34 @@
         <meta charset="utf-8">
         <meta content="width=device-width, initial-scale=1" name="viewport">
         <title>ankerctl</title>
-        <link href="/static/vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css" rel="stylesheet">
+        <link href="{{ url_for('static', filename='vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css')}}" rel="stylesheet">
+        <link href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.10.4/font/bootstrap-icons.css" rel="stylesheet">
```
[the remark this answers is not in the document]
billyjbryant: I was concerned about this, I may find an alternative for icons because pulling in all of the icons from bootstrap may be too large (IE several hundred megs just for images).

static/ankersrv.js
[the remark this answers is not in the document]
chrivers: Ah, I see what you mean - yes, the `ankersrv.js` is served by the static file handler by default.

Still, aren't these functions just copying some text that is already in the DOM?

Can't we just read the text from the element, and copy that to the clipboard? Then the issue goes away.

web/__init__.py
billyjbryant: I'm fine with removing the IF here if we are pretty sure that flask will always handle `app.post` as POST and will never accept a GET on the endpoint :)

web/__init__.py
billyjbryant: Ok, makes sense, maybe the better check here is ensuring that it is valid JSON and includes at least one of the required fields.

web/__init__.py
billyjbryant: I will test that and see how it works :)

web/__init__.py
billyjbryant: No, if you look in current code, I have forced startup to not load services if there is no login, so this would ensure services are loaded on the redirect back to `/` so that you don't have to restart the webserver.

static/ankersrv.js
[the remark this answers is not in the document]
billyjbryant: This is true, I'll re-work that and submit a new commit with the change, then all JS will be back in the `ankersrv.js` file

static/index.html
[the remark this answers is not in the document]
chrivers: I just checked the bootstrap release zip. The (uncompressed) size of the css + font versions is about 300KB, so that seems small enough to include?

web/__init__.py
chrivers: I'm fine with leaving it in, for the same reason - but let's make it early-return, instead of indenting everything :)

static/index.html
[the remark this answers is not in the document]
billyjbryant: font doesn't include icons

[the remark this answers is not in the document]
chrivers: The font from `bootstrap-icons` doesn't include icons?

web/__init__.py
chrivers: The later parsing will surely fail if the file is invalid. It'd be simpler (and good enough) to just catch exceptions from the parse step, and flash a message that the config is not valid.

No reason to try and guess if the parsing is going to work, just before doing the parsing :)

web/__init__.py
chrivers: The services are loaded on-demand. They only start up when they are requested, so the config isn't needed until then.

Also, we can't register the services twice, so this way will crash if we hit `/reload` again.

requirements.txt
```diff
@@ -10,3 +10,6 @@ crcmod==1.7
 tqdm==4.65.0
 flask==2.2.0
 flask-sock==0.6.0
+ua-parser==0.10.0
+user-agents==2.2.0
+pyyaml==5.4.1
```
chrivers, line 15: We don't seem to use pyyaml?

web/__init__.py
billyjbryant: Lines 110-114: 

```python
    login_path = {
        'Mac OS': '~/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json',
        'Windows': r'%LOCALAPPDATA%\Ankermake\AnkerMake_64bit_fp\login.json',
        'None': 'Unsupported OS, supply path to login.json',
    }
```

billyjbryant: Regarding case, I will go back and change all of the new variables to snake case instead of Camel Case.

web/__init__.py
billyjbryant: Thats not really possible here, because if you call on the config when it isn't loaded, it returns a value, it isn't a valid value but its still a positive (thus truthy) value. So using the config (or lack thereof) as an indicator for what to display using template side logic alone isn't really an option.

web/__init__.py
billyjbryant: true, I copy/pasta'd here :)

web/__init__.py
billyjbryant: Can do.

web/__init__.py
billyjbryant: Tried that, Flask threw a type error, it is expecting a set containing an array of extensions.

web/__init__.py
chrivers: Sorry, I meant `web/platform.py`, like I referenced in a later comment. But same idea :)

web/__init__.py
chrivers: That makes sense - maybe add

```python3
# required by the flash() function
app.secret...
```

requirements.txt
billyjbryant: Actually I need to remove both `ua-parser` and `pyyaml` as both are not being used in the implementation I went with. (I only ended up using `user-agents.parse` as I didn't need the full functionality of `ua-parser`).

static/index.html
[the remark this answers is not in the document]
billyjbryant: I'll double check and pull it in if that is the case :)

web/__init__.py
chrivers: This should already be the case. The `app.svc.register("foo", web.service.x.y())` lines only register the service, it doesn't activate them, until they are requested.

If we don't register the services, and then (accidentally) hit an endpoint that requires the service, we are going to get an exception anyway, because the service is missing.

Maybe a better way to handle this in general, would be a decorator. Something like `@depends_on_service("foo")`, which could then make that service available as a parameter.

In such a case, we could properly handle missing services.

But until we have that, we should register the services as normal.

If you want to see some of the logic that starts services:

https://github.com/Ankermgmt/ankermake-m5-protocol/blob/8b2a8110c03e82e01256b9d1f47c9ca8eb6d9fa5/web/lib/service.py#L261

The `ServiceManager.get()` function bumps the reference count, then does:

```python
    ....
        if self.refs[name] == 1:
            svc.start()
```

So the service gets activated when it is requested, and not before.

web/__init__.py
billyjbryant: Also, for what its worth, the `allowed_file()` function could be useful in the future (for other functions) if we are ever able to allow gcode uploads from the web interface to ensure we only get gcode :)

web/__init__.py
chrivers: A set containing an array? That doesn't sound right.

Besides, it's literally the same value:

```python3
>>> set(['json'])
{'json'}
>>> {"json"}
{'json'}
```

ankerctl.py
```diff
@@ -32,10 +32,13 @@ class Environment:
     def __init__(self):
         pass
 
-    def require_config(self):
+    def require_config(self, required=True):
         with self.config.open() as config:
             if not getattr(config, 'printers', False):
-                log.critical("No printers found in config. Please import configuration using 'config import'")
+                if not required:
```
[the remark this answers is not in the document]
chrivers: That part is fine, and reasonable. I'm just saying we should use the same error message in both cases, even if the level is different.

web/__init__.py
billyjbryant: I'm not trying to be rude, but in my testing, this is absolutely not true. In testing, in the `startup()` function, if `app.config["config"]` is not valid when the services load, it causes the app to crash and reboot. Thus I added the `if app.config["Login"]` clause to only start up those services when there is a valid login. Thus calling it in this function is calling it again once there is a valid login so the services then get registered. 

In my code, I have uploaded my config, then re-uploaded my config (which triggers `/reload` a second time) and there was no subsequent crash, the app continued to function as intended. The only ill effect i've seen in testing is that even though we don't load the video section when there is no config, it still attempts to call the pppp and mqtt service endpoints which at this point aren't registered thus triggering a 404 in logs but the end user never sees this.

ankerctl.py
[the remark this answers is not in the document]
billyjbryant: But this function is used in multiple places, if you aren't planning on using the webserver, why would we tell you to import the config via the web interface and vice-versa, if you are using the webserver and not the CLI functions, why tell you to use `config import`?

web/__init__.py
billyjbryant: So if you don't register the services and attempt to hit the endpoint you get a 404, which is an error for the end user but not an Exception and doesn't cause the app to crash. If you register the service and there is no config you DO get an Exception and the app crashes as a result, causing a restart. You can't upload the config if the app is constantly restarting itself, this was my "simple" way of solving the problem.

web/__init__.py
billyjbryant: /shrug :D I'll change it and re-test

requirements.txt
billyjbryant: No, actually I am wrong: 

```
user-agents is hosted on [PyPI](http://pypi.python.org/pypi/user-agents/) and can be installed as such:

pip install pyyaml ua-parser user-agents
```

I'll test by removing `pyyaml` and `ua-parser` but they are included as Deps of `user-agents` so removing them from `requirements.txt` shouldn't actually prevent them from being installed (since they are in the `requirements` for `user-agents`).

ankerctl.py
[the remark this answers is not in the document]
billyjbryant: ✅

static/ankersrv.js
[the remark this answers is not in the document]
billyjbryant: ✅

static/index.html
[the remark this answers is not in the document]
billyjbryant: ✅

ankerctl.py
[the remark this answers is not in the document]
chrivers: I just think it's helpful to tell the user that they have both options, since the cli and webserver currently share the same config file.

But, more than that, it seems to me that we are really forcing two different `load_config` functions into the same function body. We could also write two functions, and have one fail on missing config (`log.critical`), and the other use `log.info()` on missing config. Then it would make sense for them to have different error messages.

But this style, where one function does two different jobs depending on a parameter, is kind of confusing.

requirements.txt
chrivers: That sounds good. We should only depend on things we `import`. If those imports change their implementation to use different dependencies, that shouldn't concern us.

static/ankersrv.css
```diff
@@ -0,0 +1,21 @@
+.nav-link {
+    color: #88f387!important;
```
chrivers, line 2: Using a lot of `!important` is usually a sign that we are not quite hitting the right css scopes. Are they really required? As long as our rule has the same css specificity, and our css is loaded after bootstrap, it should override it without problems.

At the very least, please put a space before `!important` :smile:

static/ankersrv.js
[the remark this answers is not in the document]
chrivers: Good fix - much cleaner.

static/index.html
```diff
@@ -4,7 +4,13 @@
         <meta charset="utf-8">
         <meta content="width=device-width, initial-scale=1" name="viewport">
         <title>ankerctl</title>
-        <link href="/static/vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css" rel="stylesheet">
+        <link rel="stylesheet" href="{{ url_for('static', filename='vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css') }}">
+        <link rel="stylesheet" href="{{ url_for('static', filename='vendor/bootstrap-icons-1.10.5/font/bootstrap-icons.min.css') }}">
+		<link rel="stylesheet" href="{{ url_for('static', filename='ankersrv.css') }}">
```
chrivers, line 9: Listen up, tabs, we don't like your kind around here.. ;-)

static/index.html
```diff
@@ -4,7 +4,13 @@
...
+        <script type="text/javascript" src="{{ url_for('static', filename='vendor/bootstrap-5.3.0-alpha3-dist/js/bootstrap.bundle.min.js')}}"></script>
+        <script type="text/javascript" src="{{ url_for('static', filename='vendor/cash.min.js') }}"></script>
+        <script type="text/javascript" src="{{ url_for('static', filename='vendor/jmuxer.min.js') }}"></script>
+        <script type="text/javascript" src="{{ url_for('static', filename='ankersrv.js') }}"></script>
```
chrivers, line 13: Why were all `<script>` tags moved to `<head>`?

Having them in `<head>` requires all scripts to be downloaded before the browser can start rendering. It's almost never what we want.

static/vendor/bootstrap-icons-1.10.5/font/bootstrap-icons.json
chrivers, line 1: Great job on importing the bootstrap-icon bundle!

Only question is this file (`.json`). Do we use it for anything? I mean, it doesn't hurt terribly to have, but still?

web/__init__.py
```diff
@@ -98,6 +129,21 @@ def app_api_version():
     }
 
 
+@app.post('/api/ankerctl/config/upload')
+def app_api_ankerctl_config_upload():
+    with tempfile.TemporaryDirectory(prefix='ankerctl_') as tmpdir:
+        if request.method != 'POST':
+            return web.util.flash_redirect()
+        if 'login_file' not in request.files:
+            return web.util.flash_redirect('No file found', 'danger')
```
chrivers, line 138: Not quite what I meant, but looks clean. No objections :)

web/__init__.py
```diff
@@ -98,6 +129,21 @@ def app_api_version():
...
+        file = request.files['login_file']
+        if file.filename == '':
+            return web.util.flash_redirect('No file uploaded', 'danger')
```
chrivers, line 141: It's absolutely possible to upload a file without specifying a file name (quite easy mistake to make with `curl`, for example).

There's no reason to reject the upload, if the file is valid. We should just run `web.config.config_import` on it.

chrivers: This `web.config.config_import` should raise an exception if something goes wrong, and then we catch it here.

`flash_redirect()` shouldn't find its way into every corner of the codebase - ideally, we should only use it in the direct request handler functions.

web/config.py
```diff
@@ -0,0 +1,53 @@
...
+def config_import(login_file, config):
+    # extract auth token
+    cache = libflagship.logincache.load(login_file.stream.read())["data"]
+    auth_token = cache["auth_token"]
+
+    # extract account region
+    region = libflagship.logincache.guess_region(cache["ab_code"])
+
+    try:
+        newConfig = cli.config.load_config_from_api(auth_token, region, False)
```
chrivers, line 39: `newConfig`.. :)

web/config.py
```diff
@@ -0,0 +1,53 @@
...
+    except libflagship.httpapi.APIError as E:
+        message = f"Config import failed: {E} <br/> Auth token might be expired: make sure Ankermake Slicer can connect, then try again"
+        return web.util.flash_redirect(message, 'danger')
+    except Exception as E:
+        message = f"Config import failed: {E}"
+        return web.util.flash_redirect(message, 'danger')
```
chrivers, line 45: This is the pattern I'm talking about - we should move this `try-catch` logic to the request handler, and make this function just about the logic. That'll give us a chance to share the code with the cli.

web/platform.py
```diff
@@ -0,0 +1,19 @@
+def os_platform(os_family: str):
+    if os_family.startswith('Mac OS'):
+        return 'macos'
+    if os_family.startswith('Windows'):
```
chrivers, line 4: Should be `elif ... elif`, otherwise logic is fine.

web/platform.py
```diff
@@ -0,0 +1,19 @@
...
+        return 'windows'
+    if os_family.__contains__('Linux'):
```
chrivers, line 6:
```python
if os_family.__contains__('Linux'):
```

is a bad pattern. Instead:

```python
if "Linux" in os_family:
```

web/service/filetransfer.py
```diff
@@ -58,6 +57,7 @@ def handler(self, data):
             self._tap.put(msg)
 
     def worker_start(self):
+        from web import app
```
chrivers, line 60: I'm not a fan of this change.

I realize mutual imports are a bad pattern, but it's what the flask devs suggest for small or medium-sized apps, and it was manageable.

Having `import` in several functions is also pretty rough. Is there a particular reason to do it this way?

web/util.py
```diff
@@ -0,0 +1,10 @@
+from flask import flash, redirect
+
+def allowed_file(filename: str, allowed_ext: object):
```
chrivers, line 3: Are we using `allowed_file` anywhere?

Let's get rid of it.

requirements.txt
billyjbryant: This is solved ✅

web/service/filetransfer.py
billyjbryant: I was running into issues with recursive imports causing app to fail to start; however, I'm pretty sure I figured that out and will put in a new commit to fix it.

web/util.py
billyjbryant: My thought was it could be used for future upgrades (allowing uploading gcode) hence keeping it, but can get rid of it and restore it if we need it later :)

web/platform.py
billyjbryant: Fixing

.prettierrc
chrivers, line 1: Same question again - what is this file? What uses it, and why should we include it?

.vscode/settings.json
```diff
@@ -1,3 +1,8 @@
 {
-    "files.insertFinalNewline": true
+    "files.insertFinalNewline": true,
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.autopep8"
+    },
+    "python.formatting.provider": "none",
+    "python.formatting.autopep8Args": ["--ignore", "E402"]
```
chrivers, line 7: I'm not a big fan of having a git-tracked file in gitignore.

This might be a reasonable use case for that pattern. Are those settings you added good and reasonable for all vscode users on this project?

web/__init__.py
```diff
@@ -1,61 +1,83 @@
...
 app.svc = ServiceManager()
 
 sock = Sock(app)
 
-
-import web.service.pppp
-import web.service.video
-import web.service.mqtt
-import web.service.filetransfer
+import web.service.pppp  # nopep8
```
chrivers, line 33: What reacts to this syntax? If it's only vscode, please remove it.

web/__init__.py
```diff
@@ -1,61 +1,83 @@
...
+import web.service.filetransfer  # nopep8
 
 
 @app.before_first_request
 def startup():
     app.svc.register("pppp", web.service.pppp.PPPPService())
     app.svc.register("videoqueue", web.service.video.VideoQueue())
     app.svc.register("mqttqueue", web.service.mqtt.MqttQueue())
-    app.svc.register("filetransfer", web.service.filetransfer.FileTransferService())
+    app.svc.register(
```
chrivers, line 44: Is this auto-reformatting? Please undo this line's change.

web/__init__.py
```diff
@@ -80,13 +103,26 @@ def generate():
...
-    )
+    config = app.config["config"]
+    with config.open() as cfg:
+        user_agent = user_agent_parse(request.headers.get('User-Agent'))
+
+        host = request.host.split(':')
+        # If there is no 2nd array entry, the request port is 80
+        request_port = host[1] if len(host) > 1 else '80'
+        no_config = '<p>No printers found, please load your login config...</p>'
+        anker_config = str(web.config.config_show(
```
chrivers, line 114: The `anker_config` line should be a normal `if` statement. This is becoming messy.

That way we also get rid of `no_config` as a variable.

web/__init__.py
```diff
@@ -98,6 +134,20 @@ def app_api_version():
     }
 
 
+@app.post('/api/web/config/upload')
+def app_api_web_config_upload():
+    if request.method != 'POST':
+        return web.util.flash_redirect()
+    if 'login_file' not in request.files:
```
chrivers, line 141: Please add a blank line after "return" statements - it makes the structure easier to follow.

.vscode/settings.json
billyjbryant: I mean't to remove this as a git tracked file. these were settings I had in my editor but they could be useful to others (especially the autopep8 override so that it doesn't try to re-arrange imports)

.prettierrc
billyjbryant: It is for configuring prettier which is a style and formatting editor. Helps ensure that if other users have prettier in their IDE, it conforms to standards :)

.hintrc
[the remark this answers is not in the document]
billyjbryant: Controls certain VSCode (and other IDE) hints or warnings for issues.

static/ankersrv.css
```diff
@@ -0,0 +1,41 @@
...
+    color: #000000 !important;
+    background-color: #88f387 !important;
+    border: #41a03f;
+}
+
+.nav-link:hover,
+.btn:hover {
+    color: #000000;
+    background-color: #adf3ac;
+    border-color: #41a03f; /*set the color you want here*/
```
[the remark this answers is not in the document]
billyjbryant: not from stackoverflow but yes we can :D

static/index.html
```diff
@@ -1,67 +1,217 @@
...
-        </div>
-        <script src="/static/vendor/bootstrap-5.3.0-alpha3-dist/js/bootstrap.bundle.min.js"></script>
-        <script src="/static/vendor/cash.min.js"></script>
-        <script type="text/javascript" src="/static/vendor/jmuxer.min.js"></script>
-        <script type="text/javascript" src="/static/ankersrv.js"></script>
+            <script type="text/javascript" src="{{ url_for('static', filename='vendor/bootstrap-5.3.0-alpha3-dist/js/bootstrap.bundle.min.js')}}"></script>
+            <script type="text/javascript" src="{{ url_for('static', filename='vendor/cash.min.js') }}"></script>
+            <script type="text/javascript" src="{{ url_for('static', filename='vendor/jmuxer.min.js') }}"></script>
+            <script type="text/javascript" src="{{ url_for('static', filename='ankersrv.js') }}"></script>
+        </footer>
```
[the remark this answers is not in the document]
billyjbryant: actually, in html5, `<footer>` is the last element loaded in the dom model. so putting scripts in the footer is actually one of the best places because just putting them in front of the closing `</body>` doesn't ensure that they are loaded last :)

web/platform.py
chrivers, line 1: Double quotes everywhere :)

web/__init__.py
billyjbryant: it prevents vscode and any other IDE that uses pep8 for python formatting, to not move around imports all willy nilly. That is what broke the app for me that led me to moving the imports in the services. I reverted that so had to add this here to keep my IDE from fuxing me again :D

web/__init__.py
billyjbryant: Can do, and yes it was, due to line length, but I agree this makes it ugly :D

web/__init__.py
billyjbryant: Yeah, I prefer inline Ifs for variable setting but can move it to an if control block instead :)

web/__init__.py
billyjbryant: :+1:

ankerctl.py
```diff
@@ -32,10 +32,12 @@ class Environment:
     def __init__(self):
         pass
 
-    def require_config(self):
+    def load_config(self, required=True):
         with self.config.open() as config:
             if not getattr(config, 'printers', False):
-                log.critical("No printers found in config. Please import configuration using 'config import'")
+                log.warning("No printers found in config. Please upload configuration using the webserver or 'ankerctl.py config import'")
+                if required:
```
chrivers, line 39: We should use `log.critical` for this case. Something like:

```python
msg = "No printers found in config. Please upload configuration using the webserver or 'ankerctl.py config import'"
if required:
    log.critical(msg)
else:
    log.warning(msg)
```

web/config.py
billyjbryant: I'll change it to `new_config` :D

web/util.py
```diff
@@ -0,0 +1,6 @@
+from flask import flash, redirect
+
+def flash_redirect(message: str | None = None, category = 'info', path = '/'):
```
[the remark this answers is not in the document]
billyjbryant: That is how I had it first,but something in my brain made me switch the order, I'll revert it because I agree that it probably makes more sense to do it that way and not provide a default

Also, pep8 is what put the spaces around the equals, so... :D

ankerctl.py
billyjbryant: I can change that around. I was under the impression you didn't like the varying logic and wanted the message presented the same.

web/platform.py
billyjbryant: Yes, I need to set prettier to use. single quotes and reformat :D

ankerctl.py
billyjbryant: ✅

static/ankersrv.css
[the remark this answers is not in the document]
billyjbryant: ✅

web/platform.py
billyjbryant: ✔️

web/config.py
```diff
@@ -0,0 +1,49 @@
...
+        newConfig = cli.config.load_config_from_api(auth_token, region, False)
+    except libflagship.httpapi.APIError as E:
+        raise f"Config import failed: {E} <br/> Auth token might be expired: make sure Ankermake Slicer can connect, then try again"
+    except Exception as E:
+        raise f"Config import failed: {E}"
+
+    try:
+        config.save("default", newConfig)
+    except Exception as E:
+        raise f"Config import failed: {E}"
```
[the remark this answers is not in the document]
billyjbryant: ✅

web/config.py
```diff
@@ -0,0 +1,49 @@
...
+    cache = libflagship.logincache.load(login_file.stream.read())["data"]
+    auth_token = cache["auth_token"]
+
+    # extract account region
+    region = libflagship.logincache.guess_region(cache["ab_code"])
+
+    try:
+        newConfig = cli.config.load_config_from_api(auth_token, region, False)
+    except libflagship.httpapi.APIError as E:
+        raise f"Config import failed: {E} <br/> Auth token might be expired: make sure Ankermake Slicer can connect, then try again"
```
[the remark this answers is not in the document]
billyjbryant: ✅

web/config.py
billyjbryant: ✅

static/index.html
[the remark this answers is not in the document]
billyjbryant: ✅

.vscode/settings.json
spuder, line 1: I don’t think we should remove the vscode/settings file because this enforces project wide new line character consistency.

.gitignore
```diff
@@ -2,3 +2,4 @@
 __pycache__
 .DS_Store
 /settings.json
+.vscode
```
spuder, line 5: Same as below. I don’t think we should do a wholesale exclude of the vscode directory, because it is common practice to include a vscode/settings.json file for shared editor configs.

See: https://stackoverflow.com/a/57749909

static/ankersrv.css
```diff
@@ -0,0 +1,41 @@
+.nav-link,
+.btn-link {
+  color: #88f387;
+  text-decoration: none;
+}
+
+.nav-link.active,
+.btn-primary {
+  color: #000000 !important;
```
just-trey, line 9: Nit picking. `!important` should be avoided if at all possible and handled via specificity. You can leave as is and I can clean up later though when I get a chance. It does work

web/__init__.py
billyjbryant: ✅

web/__init__.py
billyjbryant: ✅

web/__init__.py
billyjbryant: ✅

web/util.py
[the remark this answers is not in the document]
billyjbryant: ✅

web/__init__.py
billyjbryant: ✅

web/__init__.py
billyjbryant: ✅

.gitignore
billyjbryant: I can revert it.

web/__init__.py
```diff
@@ -34,30 +63,63 @@
...
+    """
+    Unregisters not required services on server shutdown
+    """
+    app.svc.unregister("pppp")
+    app.svc.unregister("videoqueue")
+    app.svc.unregister("mqttqueue")
+    app.svc.unregister("filetransfer")
+
+
+def restart():
```
chrivers, line 85: This *does not work* correctly.

It may look like it does, and it might occasionally do the right thing, but it will absolutely lead to tricky bugs, and incorrect behavior.

The service manager (`app.svc`) keeps track of a number of registered services (those added with `register()`), and makes them available for use.

The web server (either endpoints, or registered services) can request access to specific services. This is tracked using reference counting.

For example, in this fragment:

```python
...
    for msg in app.svc.stream("videoqueue"):
...
```

The `.stream()` method iterates over messages from a service, and uses the service while doing so. So while streaming, the `videoqueue` service will have a reference count of at least 1.

If we `unregister()` the services, the `ServiceManager` removes the service registration, as well as the reference count to them. However, the actual `Service` objects will still be allocated, causing resource conflicts and bad behavior.

I'd argue that `unregister()` should raise an exception if the reference count is non-zero.

Since services are not activated until they are needed, all we need to do is request that the services stop, then start again.

Can you try adding this code to ServiceManager, after `unregister()`?

```python3
    def restart(self, name: str):
        if name not in self.svcs:
            raise KeyError(f"Trying to restart unknown service {name!r}")

        svc = self.svcs[name]
        svc.stop()
        svc.await_stopped()
        svc.start()
        svc.await_ready()
```

Then call `app.svc.restart()` for each service name?

It's not the most elegant solution, but maybe you can test if that works? I can cook up something more fitting, if this works.

.gitignore
chrivers: Yes, please revert.

chrivers (commented):
We're much closer - and fantastic work on the docstrings.

There's some minor nitpicks, and one serious issue about service control.

Looking forward to seeing more :)

web/__init__.py
chrivers: If you rebase on top of `chrivers/service-improvements`, there's new logic to help you out :)

The `ServiceManager` can now be iterated over, to get a list of service names.

Also, the `Service` class now has a `.restart()` function.

So something like this:

```python
for name in app.svc:
    app.svc.svcs[name].restart()
```

I'm still not convinced this is the right way to do it, but let's see if it works? :)

.vscode/settings.json
billyjbryant: I'll restore this.

.pep8
```diff
@@ -1,4 +1,4 @@
 [pycodestyle]
 max_line_length = 120
-ignore = E221,E226,E231,E241,E261,E203,E741
+ignore = E221,E226,E231,E241,E261,E203,E741,E402
```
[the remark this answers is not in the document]
billyjbryant: Per the updated description: 

> Introduced setup.cfg for editors that do not recognize .pep8 for pycodestyle formatting

setup.cfg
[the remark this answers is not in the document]
billyjbryant: > Why do we have both `.pep8` and `setup.cfg`?

Per the updated description: 

> Introduced `setup.cfg` for editors that do not recognize `.pep8` for pycodestyle formatting.

I'm using vscode with autopep8 and it was not honoring the `.pep8` file, after digging it seems that without added extensions vscode's built-in ms-autopep8 does not recognize the pep8 file but does recognize the `setup.cfg` file (which is also used by other formatters such as Black or Flake8). 

I added in the additional file for this reason. I could make one a link of the other, but sometimes links work strange cross OS and github doesn't always honor/preserve them.

.pep8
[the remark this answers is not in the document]
billyjbryant: I'll test again; however, I was still seeing modules being re-arranged without E402.

.vscode/settings.json
billyjbryant: ✔️

web/__init__.py
```diff
@@ -72,36 +134,93 @@ def ctrl(sock):
...
+    file = request.files["login_file"]
+
+    try:
+        web.config.config_import(file, app.config["config"])
+        return web.util.flash_redirect("/reload", "AnkerMake Config Imported!", "success")
+    except web.config.ConfigImportError as err:
+        log.error(err)
+        return web.util.flash_redirect("/", f"Error: {err}", "danger")
+    except Exception as err:
+        log.error(traceback.format_exc, err)
```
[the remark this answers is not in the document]
billyjbryant: ✔️

web/__init__.py
```diff
@@ -72,36 +136,93 @@ def ctrl(sock):
...
+    """
+    Returns the version details of api and server as dictionary
+
+    Returns:
+        A dictionary containing version details of api and server
+    """
+    return {"api": "0.1", "server": "1.9.0", "text": "OctoPrint 1.9.0"}
+
+
+@app.post("/api/web/config/upload")
```
chrivers, line 193: As I mentioned, we should have `/ankerctl/` in the api scope for our internal endpoints. For example, `/api/ankerctl/config/upload` for this endpoint.

Since `/api` is used for compatibility with OctoPrint, we have to be quite sure we don't have a name clash with OctoPrint, and `/web` is certainly generic enough that it might be used.

We could also move our scope somewhere else, since it's entirely under our control, but it would make sense to keep it under `/api`, since that might very well be the authenticated scope if we later add a login system.

So please change to `/api/ankerctl/config/upload`, or other reasonable non-clashing name.

(same kind of change needed for other ankerctl-specific endpoints)

web/__init__.py
```diff
@@ -117,9 +238,48 @@ def app_api_files_local():
     return {}
 
 
+@app.get("/reload")
```
chrivers, line 241: `/api/ankerctl/server/reload` maybe?

.pep8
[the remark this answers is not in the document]
chrivers: Is it because vscode does not support the `# autopep8: {off,on}` syntax? Or did it end up working?

setup.cfg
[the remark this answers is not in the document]
chrivers: I can see the problem - it seems like you found a solution for getting vscode to recognize the `.pep8` file, so this should be good for now :+1: 

I agree that symlinks are tricky, and not ideal.

.vscode/settings.json
```diff
@@ -1,3 +1,12 @@
 {
-    "files.insertFinalNewline": true
+    "files.insertFinalNewline": true,
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.autopep8",
+        "editor.formatOnSave": true,
```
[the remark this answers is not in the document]
billyjbryant: So with the changes I have configured, it will format on save but only format the portions of the file that changed. Additionally, some of the changes you indicate occurred BEFORE we settled on `pep8` as a linting and formatting engine and standard for the project. Introducing `pep8` with your desired config won't go back and re-write those changes. 

I included : `        "editor.formatOnSaveMode": "modifications"` in the vscode settings so that it only changes sections that were modified before the save that way it reduces future PR issues where the editor goes nuts reformatting. Keep in mind, this ONLY effects VSCode, no other editor is going to honor this setting. 

Regarding:
```
    "python.linting.pycodestyleArgs": ["--config", ".pep8"],
    "python.formatting.autopep8Args": ["--config", ".pep8"],
```

These were added to tell VSCode where it should get its `pycodestyle` settings from as by default `pycodestyle` and `autopep8` look for them in `setup.cfg`. I introduced these lines to eliminate the need for a redundant config file to achieve the same purpose.

web/__init__.py
billyjbryant: Implemented and seems to work better.

web/__init__.py
billyjbryant: Do you also want `/reload` to become `/ankerctl/reload`?

.vscode/settings.json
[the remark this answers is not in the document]
chrivers: Okay, I thought those where manual changes. Thanks for the explanation - everything sounds reasonable. Let's go with this version.

web/__init__.py
```diff
@@ -34,30 +64,58 @@
...
 
 
+def restart():
+    """
+    Restarts the registered services to apply new changes
+    """
+    try:
+        for name in app.svc:
+            app.svc.svcs[name].restart()
+            time.sleep(2)
```
chrivers, line 83: Wow, why are we doing a sleep between each service restart?

If this is needed, there's a bug we need to fix.

web/__init__.py
```diff
@@ -34,30 +64,58 @@
...
+    except Exception as err:
+        log.exception(err)
+        return Exception(err)
```
chrivers, line 86: Are we returning exceptions now?

Since the try-except is on the outside of the loop, this changes almost nothing, compared to not having the loop.

web/__init__.py
billyjbryant: ✅

.gitignore
billyjbryant: Done

web/__init__.py
billyjbryant: In my testing, the timing of the different services restarting was causing this function to exit before the services had stopped and restarted successfully thus reloaded to a page where there were no running services resulting in a 500 error.

web/__init__.py
billyjbryant: it is to return the exception back to the function that called it as an exception instead of a return message.

web/__init__.py
chrivers: can you try this pattern:

```python
   for name in app.svc:
       app.svc.svcs[name].restart()

   for name in app.svc:
       app.svc.svcs[name].await_ready()
```

billyjbryant: Ok, so I tried that and it got rid of the 500 exception I was seeing previously; however, it resulted in the site redirecting and the video feed to be broken for several seconds. 

My mechanism for testing this was: 
1. Open browser to http://127.0.0.1:4470
1. Navigate to **Setup** tab
1. Click on the `Restart Services` button

<details><summary>Expand log</summary>

```py
❯ python ./ankerctl.py webserver run
 * Serving Flask app 'web'
 * Debug mode: off
[*] WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:4470
[*] Press CTRL+C to quit
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET / HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/vendor/bootstrap-icons-1.10.5/font/bootstrap-icons.min.css HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/ankersrv.css HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/vendor/bootstrap-5.3.0-alpha3-dist/js/bootstrap.bundle.min.js HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/img/setup/prusaslicer-2.png HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/img/setup/prusaslicer-3.png HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/img/setup/prusaslicer-1.png HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/vendor/cash.min.js HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/vendor/jmuxer.min.js HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/ankersrv.js HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/img/setup/prusaslicer-4.png HTTP/1.1" 200 -
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /static/vendor/bootstrap-icons-1.10.5/font/fonts/bootstrap-icons.woff2?1fa40e8900654d2863d011707b9fb6f2 HTTP/1.1" 200 -
[*] MqttQueue: Requesting start
[*] Connecting to make-mqtt.ankermake.com
[*] VideoQueue: Requesting start
[*] PPPPService: Requesting start
[*] Trying connect over pppp
[*] 127.0.0.1 - - [10/May/2023 14:01:31] "GET /favicon.ico HTTP/1.1" 404 -
[*] Established pppp connection
[*] PPPPService: Worker started
[*] VideoQueue: Worker started
[*] Connected to mqtt
[*] MqttQueue: Worker started
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] PPPPService: Requesting restart
[*] PPPPService: Requesting stop
[*] PPPPService: Worker stopped
[*] PPPPService: Requesting start
[*] Trying connect over pppp
[*] VideoQueue: Service requested restart.
[*] Established pppp connection
[*] PPPPService: Worker started
[*] VideoQueue: Requesting restart
[*] VideoQueue: Requesting stop
[E] CLOSE
[*] PPPPService: Service requested restart.
[*] PPPPService: Requesting stop
[*] VideoQueue: Worker stopped
[*] VideoQueue: Requesting start
[*] PPPPService: Requesting start
[*] PPPPService: Worker stopped
[*] Trying connect over pppp
[*] Established pppp connection
[*] PPPPService: Worker started
[*] VideoQueue: Worker started
[*] MqttQueue: Requesting restart
[*] MqttQueue: Requesting stop
[*] MqttQueue: Worker stopped
[*] MqttQueue: Requesting start
[*] Connecting to make-mqtt.ankermake.com
[*] Connected to mqtt
[*] MqttQueue: Worker started
[*] FileTransferService: Requesting restart
[*] FileTransferService: Requesting stop
[E] FileTransferService: Waiting for stopped thread
Traceback (most recent call last):
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/web/__init__.py", line 85, in restart
    app.svc.svcs[name].await_ready()
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/web/lib/service.py", line 206, in await_ready
    raise ServiceStoppedError(f"{self.name}: Waiting for stopped thread")
web.lib.service.ServiceStoppedError: FileTransferService: Waiting for stopped thread
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /api/ankerctl/server/reload HTTP/1.1" 302 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET / HTTP/1.1" 200 -
[*] VideoQueue: Requesting stop
[*] PPPPService: Requesting stop
[*] VideoQueue: Worker stopped
[*] PPPPService: Worker stopped
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/vendor/bootstrap-icons-1.10.5/font/bootstrap-icons.min.css HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/img/setup/prusaslicer-1.png HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/ankersrv.css HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/img/setup/prusaslicer-2.png HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/img/setup/prusaslicer-3.png HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/img/setup/prusaslicer-4.png HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/vendor/cash.min.js HTTP/1.1" 304 -
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/vendor/jmuxer.min.js HTTP/1.1" 304 -
[*] MqttQueue: Requesting stop
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/vendor/bootstrap-5.3.0-alpha3-dist/js/bootstrap.bundle.min.js HTTP/1.1" 304 -
[*] 127.0.0.1 - - [10/May/2023 14:01:38] "GET /static/ankersrv.js HTTP/1.1" 304 -
[*] MqttQueue: Requesting start
[*] VideoQueue: Requesting start
[*] PPPPService: Requesting start
[*] Trying connect over pppp
[*] Established pppp connection
[*] PPPPService: Worker started
[*] VideoQueue: Worker started
[E] CLOSE
[*] PPPPService: Service requested restart.
[*] PPPPService: Worker stopped
[*] Trying connect over pppp
[*] VideoQueue: Service requested restart.
[E] CLOSE
[E] PPPPService: Failed to start worker: . Retrying in 1 second.
Traceback (most recent call last):
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/web/lib/service.py", line 102, in _attempt_start
    self.worker_start()
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/web/service/pppp.py", line 46, in worker_start
    api.process(msg)
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/libflagship/ppppapi.py", line 300, in process
    raise ConnectionResetError
ConnectionResetError
[*] PPPPService: Requesting stop
[*] VideoQueue: Worker stopped
[*] PPPPService: Requesting start
[*] Trying connect over pppp
[E] CLOSE
[E] PPPPService: Failed to start worker: . Retrying in 1 second.
Traceback (most recent call last):
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/web/lib/service.py", line 102, in _attempt_start
    self.worker_start()
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/web/service/pppp.py", line 46, in worker_start
    api.process(msg)
  File "/Users/billybryant/github/personal/ankermake-m5-protocol/libflagship/ppppapi.py", line 300, in process
    raise ConnectionResetError
ConnectionResetError
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] Trying connect over pppp
[*] Established pppp connection
[*] PPPPService: Worker started
[*] VideoQueue: Worker started
[E] CLOSE
[*] PPPPService: Service requested restart.
[*] PPPPService: Worker stopped
[*] Trying connect over pppp
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] Established pppp connection
[*] PPPPService: Worker started
[*] VideoQueue: Service requested restart.
[*] PPPPService: Requesting stop
[*] VideoQueue: Worker stopped
[*] PPPPService: Requesting start
[*] VideoQueue: Worker started
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
```

</details>

web/__init__.py
billyjbryant: Yes sorry, it should be `raise` not `return` that is my own mixup, I have now eliminated the try-except here and moved the log functionality into the reload command.

billyjbryant: Doing that actually causes the exception encountered when the `filetransfer` service fails to stop to pass to the front-end and emit a failure in reloading the services. if you wait a few seconds then refresh the page the video is then reloaded

web/__init__.py
chrivers: As long as it recovers correctly (which it seems to do), I think that's good enough for now.

I want to implement actual dependency control in the ServiceManager, which would resolve this problem.

It's complicated to track the dynamic relationships between the services, but the version we have now is better than an arbitrary sleep - which *will* occasionally fail. So let's go with what we have now :)

web/__init__.py
chrivers: Great - looks much simpler now :)

web/__init__.py
chrivers: Yes - and I can see it's already done. Great :+1:

billyjbryant:
We good to approve and merge this @chrivers ? :)

chrivers (approved):
Okay, let's do this :sunglasses:

PR #54, opened 2023-05-04
Chrivers/pre-pep8-prep

chrivers:


.pep8
```diff
@@ -0,0 +1,4 @@
+[pycodestyle]
+max_line_length = 120
+ignore = E221,E226,E231,E241,E261,E203,E741
+exclude = libflagship/pppp.py,libflagship/mqtt.py,libflagship/amtypes.py
```
chrivers, line 4: In case anyone's wondering: These files are excluded, because they are auto-generated from the protocol spec files.

chrivers:
Looks good to me, the author.

Issue #53, opened 2023-05-02
Multiple Printer support


just-trey:
Currently this solution only supports one printer. If multiple printers are detected it chooses the first one in the array. I would like a solution that supports multiple printers or at least let's us specify one to connect to.

kivinen:
I needed this myself as my friend bought ankermake too, and now I have two printers configured. 

There is pull request #55 to implement that.

This is minimal implementation i.e., it refers printers using their indexes, thus saying 

`python3 ankerctl.py -p 1 pppp capture-video -m 1m output.h264`

or 

`python3 ankerctl.py -p 1 webserver run`

to select printer number 1, i.e., your second printer.

billyjbryant:
fixed by #55

thomasjpatterson:
Ready to Ship

PR #55, opened 2023-05-05
Added support for selecting printer using -p option.

kivinen:
This is needed if you have multiple printers. Printers are identified by the index in the array, i.e., first printer is printer 0, next is printer 1.


billyjbryant (commented):
We were just discussing this the other night in the discord. This is a nice change; however, maybe we should also add a way of listing the printers available so you know what index number to use, either that or also accept a printer's S/N as input.

billyjbryant:
Can you also add `fixes #53` to your issue description as a new line? This will link the PR to the Issue to auto resolve it once merged.

kivinen:
Billy Bryant writes:
> @billyjbryant commented on this pull request.
> 
> We were just discussing this the other night in the discord. This is a nice
> change; however, maybe we should also add a way of listing the printers
> available so you know what index number to use, either that or also accept a
> printer's S/N as input.

I already changed the config show to list printer numbers:

*] Loading config from /home/kivinen/.config/ankerctl/default.json

[*] Account:
    user_id:    XXXX...<REDACTED>
    auth_token: XXXX...<REDACTED>
    email:      ***@***.***
    region:     EU

[*] Printers:
    printer:   0
    duid:      EUXXXX-00YYYY-AAAAA
    sn:        AKXXXXXXXX
    ip:        192.168.4.22
    wifi_mac:  E8:EE:CC:ZZ:YY:XX
    api_hosts: 18.168.14.146, 15.188.180.60, 18.157.241.124
    p2p_hosts: 18.157.241.124, p2p-mk-lon.ankermake.com, p2p-mk-par.ankermake.com

    printer:   1
    duid:      EUXXXX-00ZZZZ-AAAAA
    sn:        AKXXXXXXXXX
    ip:        10.21.0.226
    wifi_mac:  E8:EE:CC:XX:YY:ZZ
    api_hosts: 18.157.241.124, 18.168.14.146, 15.188.180.60
    p2p_hosts: 18.157.241.124, p2p-mk-lon.eufylife.com, p2p-mk-par.eufylife.com

I.e., you can see the printers there. I also added some of the debug
prints to tell you to which printer you are connecting to. 
-- 
***@***.***

just-trey:
If this gets done

just-trey:
I also have opened #53 and this would be helpful for that.

cli/mqtt.py
```diff
@@ -13,10 +13,13 @@
 
 def mqtt_open(env):
     with env.config.open() as cfg:
-        printer = cfg.printers[0]
+        if env.printer >= len(cfg.printers):
```
chrivers, line 16: It's great that we're doing error checking, but we shouldn't be duplicating code (between here and the pppp couterpart).

There's an easy fix! The `env` is of type `Environment` (see the top of this file), and it's made for exactly this kind of task.

Look at the comment for that class, to see how to extend it.

After it's fixed, we can simply refer to the printer through `env` :)

cli/mqtt.py
```diff
@@ -13,10 +13,13 @@
...
+            log.fatal(f"Printer number {env.printer} out of range, max printer number is {len(cfg.printers)-1} ")
+            return
```
chrivers, line 18: This pattern (fatal log + return) has an easier pattern in this project: `log.critical()` exits the program with an error message (this is a project-specific override)

cli/pppp.py
```diff
@@ -13,10 +13,13 @@
...
     with env.config.open() as cfg:
-        printer = cfg.printers[0]
+        if env.printer >= len(cfg.printers):
+            log.fatal(f"Printer number {env.printer} out of range, max printer number is {len(cfg.printers)-1} ")
+            return
+        printer = cfg.printers[env.printer]
 
         api = AnkerPPPPApi.open_lan(Duid.from_string(printer.p2p_duid), host=printer.ip_addr)
-        log.info("Trying connect over pppp")
+        log.info(f"Trying connect to printer {printer.p2p_duid} over pppp using ip {printer.ip_addr}")
```
chrivers, line 22: Excellent improvement - definitely much more informative :+1:

ankerctl.py
```diff
@@ -439,13 +443,15 @@ def config_show(env):
         print()
 
         log.info("Printers:")
-        for p in cfg.printers:
+        for i, p in enumerate(cfg.printers):
```
chrivers, line 446: There's a slight problem here - because we are just iterating over `cfg.printers`, we need to make sure the ordering is consistent, to have any meaningful correlation between the index, and the intended printer.

In Python 3, the default dict type is now (if I remember correctly), always using ordered keys. So there will be some kind of correlation, but it's mostly luck.

To really make this work, we need to sort the printer list in a predictable way. For example, by serial, or `duid`.

Otherwise, we might get a new random order, every time the printer list is imported from the API, and this can happen automatically (e.g., when upgrading ankerctl).

It would be quite bad to have your printer indices randomly shuffled in this event, so we should avoid it.

Do you want some tips on how to do this, or would you like to give it a go? :)

ankerctl.py
```diff
@@ -81,6 +82,9 @@ def main(ctx, verbose, quiet, insecure):
 
     env.upgrade_config_if_needed()
 
+    env.printer = printer
+    log.debug(f"Using printer [{env.printer}]");
```
chrivers, line 86: Naming-wise, using the word `printer` to refer to an *index* is a little confusing, since everywhere else in the code, we use `printer` to refer to a printer-related structure (such as a specific printer config).

I'd be ok with just `index`, if you want to keep it short, but otherwise something like `printer_index`? Suggestions welcome.

ankerctl.py
```diff
@@ -81,6 +82,9 @@ def main(ctx, verbose, quiet, insecure):
 
     env.upgrade_config_if_needed()
 
+    env.printer = printer
```
chrivers, line 85: We should add a method to the `Environment` class to handle this, instead of patching the value into it: `def select_printer(self, index)` on `Environment`.

This function should do the following:
 1. Check that the index is within range
 2. Open the config, and sort the list in a well-defined way (see later comment on `enumerate`)
 3. Save this list to `self._printers`
 4. Save the index to `self._index`

Then, we can add a property:

```python
@property
def printer(self):
    return self._printers[self._index]
```

This way, we can use `env.printer` to refer to the selected printer, anywhere we need it. This should make the whole thing simpler.

chrivers (changes requested):
Hello kivinen!

Welcome to the project, and thank you so much for your contribution :)

You've implemented a feature that is sorely needed, and I'm excited to see `ankerctl` being able to do this!

I have a few comments and concerns before we can merge - adressing these will keep the project in top shape going forward. If you need any help, or have any questions, please let me know.

Looking forward to seeing the next iteration of this.

kivinen:
Christian Iversen writes:
> In cli/mqtt.py:
> 
> > @@ -13,10 +13,13 @@
>  
>  def mqtt_open(env):
>      with env.config.open() as cfg:
> -        printer = cfg.printers[0]
> +        if env.printer >= len(cfg.printers):
> 
> It's great that we're doing error checking, but we shouldn't be
> duplicating code (between here and the pppp couterpart).
> 
> There's an easy fix! The env is of type Environment (see the top of
> this file), and it's made for exactly this kind of task.
> 
> Look at the comment for that class, to see how to extend it.
> 
> After it's fixed, we can simply refer to the printer through env :)

To do the checking in the Environment class would require it to open
config every time the index is needed, which would be quiet
inefficient, when the pppp_open first opens config, then fetches
printer_index from the env which will again open config etc.

Perhaps we should instead add class to Environment:

    def current_printer(self):
        with self.config.open() as config:
	    if self.printer_index >= len(config.printers):
	       log.critical("Printer number {env.printer} out of range, max printer number is {len(config.printers)-1} ")
	    return config.printers[self.printer]

or something like that. Then the pppp and mqtt open would simply say

    printer = env.current_printer()

On the other hand the mqtt also wants to get the account information
from the config, so we might need a function in Environment to return
config (and perhaps cache it there after it has been opened) or
something.



> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In cli/mqtt.py:
> 
> > @@ -13,10 +13,13 @@
>  
>  def mqtt_open(env):
>      with env.config.open() as cfg:
> -        printer = cfg.printers[0]
> +        if env.printer >= len(cfg.printers):
> +            log.fatal(f"Printer number {env.printer} out of range, max printer number is {len(cfg.printers)-1} ")
> +            return
> 
> This pattern (fatal log + return) has an easier pattern in this project: log.critical() exits the program with
> an error message (this is a project-specific override)

Ok.

> In ankerctl.py:
> 
> > @@ -439,13 +443,15 @@ def config_show(env):
>          print()
>  
>          log.info("Printers:")
> -        for p in cfg.printers:
> +        for i, p in enumerate(cfg.printers):
> 
> There's a slight problem here - because we are just iterating over
> cfg.printers, we need to make sure the ordering is consistent, to
> have any meaningful correlation between the index, and the intended
> printer.
> 
> In Python 3, the default dict type is now (if I remember correctly),
> always using ordered keys. So there will be some kind of
> correlation, but it's mostly luck.
> 
> To really make this work, we need to sort the printer list in a
> predictable way. For example, by serial, or duid.
> 
> Otherwise, we might get a new random order, every time the printer
> list is imported from the API, and this can happen automatically
> (e.g., when upgrading ankerctl).
> 
> It would be quite bad to have your printer indices randomly shuffled
> in this event, so we should avoid it.
> 
> Do you want some tips on how to do this, or would you like to give
> it a go? :)

Sorting them by serial or duid is not a good idea, as the list can
change then too, i.e., if you happen to add new printer that has lower
serial number or duid than your previous printer then the -p 0 is not
same printer it used to be.

Best would be to add some kind of nickname for printers that is stored
in ankerctl local config and use that instead of index. I used index
as quick hack to get it working, as adding names for printer requires
more changes.

I was tryign to look what the rename-printer does, but it seems to
simply send some mqtt command to cloud and I do know if that name is
visible anywhere.


> In ankerctl.py:
> 
> > @@ -81,6 +82,9 @@ def main(ctx, verbose, quiet, insecure):
>  
>      env.upgrade_config_if_needed()
>  
> +    env.printer = printer
> +    log.debug(f"Using printer [{env.printer}]");
> 
> Naming-wise, using the word printer to refer to an index is a little
> confusing, since everywhere else in the code, we use printer to
> refer to a printer-related structure (such as a specific printer
> config).
> 
> I'd be ok with just index, if you want to keep it short, but
> otherwise something like printer_index? Suggestions welcome.

I think that in the long run we want to change from printer indexes to
user settable names, and perhaps we should use printer_name or
something like that already now. 

> In ankerctl.py:
> 
> > @@ -81,6 +82,9 @@ def main(ctx, verbose, quiet, insecure):
>  
>      env.upgrade_config_if_needed()
>  
> +    env.printer = printer
> 
> We should add a method to the Environment class to handle this, instead of patching the value into it: def
> select_printer(self, index) on Environment.

I was just copying from env.config, env.insecure, env.level, and
env.log...

> This function should do the following:
> 
>  1. Check that the index is within range
>  2. Open the config, and sort the list in a well-defined way (see later comment on enumerate)
>  3. Save this list to self._printers
>  4. Save the index to self._index

To verify that the index is within range would require opening config,
and we might not yet have config available (for example during config
import), so we might not be able to open config. I think the actual
range checking should be done when the printer_index/printer_name is
used. I think making user settable aliases / names for printers is
better option than trying to make stable indexes for them. 

> 
> Then, we can add a property:
> 
> @property
> def printer(self):
>     return self._printers[self._index]
> 
> This way, we can use env.printer to refer to the selected printer, anywhere we need it. This should make the
> whole thing simpler.
-- 
***@***.***

chrivers:
You raise some good points - I'll get back to you when I can (in about a day or so)

billyjbryant:
Would it be possible to add an Environment variable that can be used to also provide the printer number? This would be useful for those running Ankerctl in Docker to provide the printer number by ENV rather than having to edit or override the start commands. 

Something like PRINTER=0 or PRINTER=1 that can be used instead of the command line parameters.

billyjbryant:
> Sorting them by serial or duid is not a good idea, as the list can
change then too, i.e., if you happen to add new printer that has lower
serial number or duid than your previous printer then the -p 0 is not
same printer it used to be.
> 
> Best would be to add some kind of nickname for printers that is stored
in ankerctl local config and use that instead of index. I used index
as quick hack to get it working, as adding names for printer requires
more changes.

I think the best constant we can use is the printer's serial number, which is exposed to the user in multiple areas (App, Anker Slicer, Printer Interface), we can take a case-insensitive input of the serial number here and that solves the issue, you select the printer which has the right serial number and there is no need to care about index order. It should be relatively trivial. 

> I was tryign to look what the rename-printer does, but it seems to
simply send some mqtt command to cloud and I do know if that name is
visible anywhere.

Anker does not seem to store the printer name/alias in any of the local configs and I'm not certain if it is ever exposed by any interface to the authentication broker. I think relying on a name would be great, but only if we can pull the one set by Anker, that way it is a reliable piece of information that remains constant so long as the user doesn't change it in the Anker app.

kivinen:
Billy Bryant writes:
> Would it be possible to add an Environment variable that can be used
> to also provide the printer number? This would be useful for those
> running Ankerctl in Docker to provide the printer number by ENV
> rather than having to edit or override the start commands.
> 
> Something like PRINTER=0 or PRINTER=1 that can be used instead of
> the command line parameters.

Yes, I was thinking that something like that might be useful, but
didn't implement that in this first version. Unfortunately I am going
to be very busy next week, as I have to be in IEEE conference week
after that thus I do not have any time to work on this code until I
get back, i.e. around end of May.
-- 
***@***.***

billyjbryant:
> Unfortunately I am going to be very busy next week, as I have to be in IEEE conference week after that thus I do not have any time to work on this code until I get back, i.e. around end of May.

If you'd like, I can make the modifications and submit a PR to your branch to merge it in.

billyjbryant:
@kivinen : https://github.com/kivinen/ankermake-m5-protocol/pull/1

billyjbryant:
fixes: #53

kivinen:
Billy Bryant writes:
>     Unfortunately I am going to be very busy next week, as I have to be in
>     IEEE conference week after that thus I do not have any time to work on
>     this code until I get back, i.e. around end of May.
>    
> If you'd like, I can make the modifications and submit a PR to your branch to
> merge it in.

Sure, go ahead.
-- 
***@***.***

billyjbryant:
@kivinen two things:

1. I submitted https://github.com/kivinen/ankermake-m5-protocol/pull/1 which adds the ability to specify the printer index via an environment variable
2. Could you base this off of [upcoming/v1.0.0](https://github.com/Ankermgmt/ankermake-m5-protocol/tree/upcoming/v1.0.0) instead of [master](https://github.com/Ankermgmt/ankermake-m5-protocol/tree/master), this is the branch that is being prepared to be merged in and has a lot of major changes.

rjchrhl:
The printer's user-assignable name appears to be "station_name" from the JSON response from appapi.query_fdm_list()

For example, for my two M5s:

Printer #0:  station_name = "Brass AnkerMake"
Printer #1:  station_name = "Hardened Steel AnkerMake"

It would be great if those names could be used to reference each individual printer.

billyjbryant:
> It would be great if those names could be used to reference each individual printer.

We could absolutely add that field into the interface and the output of `config show`, I don't think it works for selecting a printer though, especially for long names like what you have

ankerctl.py
billyjbryant: I added in functionality to sort by printer.id here, this is the id assigned by AnkerMake. Yes, this sort order could change if a printer is ingested with a lower id, but so can any other field that we choose to sort by. I think sorting by printer id makes the most sense as it is a constant that should never change and likely will only ever increase rather than decrease, thus the risk of the sort order getting messed up is likely reduced.

billyjbryant:
> Anker does not seem to store the printer name/alias in any of the local configs and I'm not certain if it is ever exposed by any interface to the authentication broker.

I was wrong on this, the name of the printer is stored in the `station_name` field and this is now being exposed in config.

ankerctl.py
billyjbryant: Updated this to `printer_index`

cli/mqtt.py
billyjbryant: If we view the efficiency loss of having to import the config multiple times if/when necessary to be negligible, I can go ahead and implement the requested change. I also like the idea of potentially caching the config and referencing it from cache when needed.

cli/pppp.py
billyjbryant: I expanded upon this by adding in the printer's name to this output.

billyjbryant:
@chrivers this pr is ready for re-review

ankerctl.py
chrivers: That makes sense - and I agree, there's no 100% solution for this, that's simply impossible.

We could consider using the `create_time`, since that reflects a timestamp related to printer registration (not sure if it's when the printer was first registered, or most recently, but it should be one of those).

I'm also perfectly happy with station_id - as long as we chose something predictable, like you say.

Comments?

cli/model.py
```diff
@@ -61,4 +65,4 @@ def mqtt_password(self):
 @dataclass
 class Config(Serialize):
     account: Account
-    printers: [Printer]
+    printers: list[Printer]
```
chrivers, line 68: My bad - good catch :+1:

ankerctl.py
```diff
@@ -440,20 +444,29 @@ def config_show(env):
...
+        print(f"    user_id:    {cfg.account.user_id[:10]}...<REDACTED>")
+        print(f"    auth_token: {cfg.account.auth_token[:10]}...<REDACTED>")
         print(f"    email:      {cfg.account.email}")
         print(f"    region:     {cfg.account.region.upper()}")
         print()
 
         log.info("Printers:")
-        for p in cfg.printers:
+        # Sort the list of printers by printer.id
+        cfg.printers.sort(key=lambda p: p.id)
```
chrivers, line 455: I just realized that we don't need to sort here.

When we import the printer list from the http api (in `cli/config.py`), we do get the list back unsorted, and so we need to sort that.

But this value comes from our own config file, which is stored as a list of dictionaries, and as such the order is preserved.

Let's remove this sort, it's a NOP :)

chrivers (commented):
Looking good - just a few minor questions, but we are almost ready to merge this :)

chrivers:
Oh, and this needs a rebase against the `upcoming/v1.0.0` branch :)

ankerctl.py
billyjbryant: > We could consider using the `create_time`, since that reflects a timestamp related to printer registration (not sure if it's when the printer was first registered, or most recently, but it should be one of those).

I tested this by de-registering my printer and re-registering it and it updated the `create_time`, so it appears to be "registration time" but is related to the most recent registration, not the first registration. 



> I'm also perfectly happy with station_id - as long as we chose something predictable, like you say.

I think `station_id` should be fairly static, during my de-registration/re-registration testing, it didnt change while `create_time` did. For that reason, I think it makes perfect sense to use this as the sorting key.

web/util.py
```diff
@@ -24,3 +26,30 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
...
+    Returns:
+    dict: Dictionary containing sanitized printer information.
+
+    The function logs warnings if the provided index is out of range and defaults it to 0. 
+    It removes 'mqtt_key' and 'p2p_key' from the printer information before returning.
+    """
+    with config.open() as cfg:
+        printers = cfg.printers
+        if index >= len(printers):
+            log.warning(f"Printer number {index} out of range, max printer number is {len(printers)-1}")
```
billyjbryant, line 48: I'm iffy on this particular portion, part of me thinks it isn't necessary here because we shouldn't get to this point with an invalid printer index, but then again, if someone were to re-use this function, they could set a static index that could fall out of range. Thoughts?

web/util.py
```diff
@@ -24,3 +26,30 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
...
+            log.warning(f"Defaulting index to 0")
+            index = 0
+
+        printer_dict = asdict(printers[index])
+        printer_dict.pop("mqtt_key", None)
+        printer_dict.pop("p2p_key", None)
```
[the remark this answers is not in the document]
billyjbryant: > As far as I can tell, the result of `get_printer()` is only used in one place, by `render_template()`. So if we just don't refer to these values in the template, all should be good, right?

I'm not as familiar with flask and was unaware as to whether or not the entire dataset would be accessible through the developer's console even if flask didn't access it on the render. Thus the security engineer in me said "Sanitize the data to remove sensitive fields"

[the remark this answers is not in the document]
chrivers: Ah, gotcha. That makes sense, for sure.

But in this case, we are doing pure server-side rendering, so we don't have to worry about that (as long as we are controlling the templates).

So that should make things easier :)

static/index.html
```diff
@@ -143,11 +143,11 @@
                 <div class="container">
                     <div class="row align-items-start">
                         <div class="col-lg-8">
-                            {%if printer_serial %}
+                            {%if printer %}
                             <div class="row mb-3">
                                 <div class="container rounded-3" id="connection">
                                     <div class="p-3 rounded-3 fs-6 bg-body-tertiary">
-                                        Connected to: <strong>{{ printer_serial }}</strong>
+                                        Connected to: <strong>{{ printer.name }} ({{ printer.sn }})</strong>
```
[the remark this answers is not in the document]
billyjbryant: Tested this by passing a javascript injection: 

image
image

This was tested both with and without jinja2 escaping with `| e`. I think it is safe to say that we are safe from injection, nevertheless, I am going to implement the escape anyway.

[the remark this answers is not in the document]
billyjbryant: ✅

ankerctl.py
billyjbryant: ✅

web/config.py
```diff
@@ -46,10 +47,17 @@ def config_show(config: object):
         region:     {config.account.region.upper()} </p
     """
     config_output += "<p>Printers:</p><hr/>"
-    for printer in config.printers:
+    for i, printer in enumerate(config.printers):
+        create_time = datetime.fromtimestamp(printer.create_time)
```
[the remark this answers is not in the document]
billyjbryant: ✅

billyjbryant (commented):
@chrivers should be good to go now with all changes we discussed.

billyjbryant:
Testing last change : 

```bash
billy  …/ankermake-m5-protocol   multiprinter-support $ via 🐍 v3.10.6 took 22s
 󱑇 00:11 󰬡    PRINTER_INDEX=1 FLASK_PORT=4480 ./ankerctl.py webserver run
[!] Printer number 1 out of range, max printer number is 0
[*] ServiceManager: Shutdown complete
```

cli/config.py
```diff
@@ -107,10 +108,17 @@ def load_config_from_api(auth_token, region, insecure):
...
+    # Sort the list of printers by printer.id
+    printers.sort(key=lambda p: p["station_id"])
     for pr in printers:
         station_sn = pr["station_sn"]
         config.printers.append(Printer(
+            id=pr["station_id"],
             sn=station_sn,
+            name=pr["station_name"],
+            model=pr["station_model"],
+            create_time=pr["create_time"],
```
chrivers, line 120: There's a subtle bug here, which doesn't manifest in our current code.

The config objects need to be structured data (so, `datetime()` instance instead of timestamp).

Here, we are creating a `Printer` object, where `.create_time` and `.update_time` are timestamps, not `datetime` objects.

So the return value from `load_config_from_api()` isn't a valid config, but because we happen to save and then load it, the timestamp is parsed when loading. So we mask that error. But if we had implemented, say, caching, we could have hit some very strange errors now :smile: 

Long story short, please apply this patch:


```diff
diff --git cli/config.py cli/config.py
index bd93525..0b0e116 100644
--- cli/config.py
+++ cli/config.py
@@ -117,8 +117,8 @@ def load_config_from_api(auth_token, region, insecure):
             sn=station_sn,
             name=pr["station_name"],
             model=pr["station_model"],
-            create_time=pr["create_time"],
-            update_time=pr["update_time"],
+            create_time=datetime.fromtimestamp(pr["create_time"]),
+            update_time=datetime.fromtimestamp(pr["update_time"]),
             mqtt_key=unhex(pr["secret_key"]),
             wifi_mac=pr["wifi_mac"],
             ip_addr=pr["ip_addr"],
diff --git cli/model.py cli/model.py
index 0cdaea7..6e9631c 100644
--- cli/model.py
+++ cli/model.py
@@ -23,6 +23,8 @@ def to_dict(self):
             res[k] = getattr(self, k)
             if v.type == bytes:
                 res[k] = enhex(res[k])
+            elif v.type == datetime:
+                res[k] = res[k].timestamp()
         return res

     @classmethod
```

static/index.html
```diff
@@ -143,11 +143,11 @@
                 <div class="container">
                     <div class="row align-items-start">
                         <div class="col-lg-8">
-                            {%if printer_serial %}
+                            {%if printer %}
                             <div class="row mb-3">
                                 <div class="container rounded-3" id="connection">
                                     <div class="p-3 rounded-3 fs-6 bg-body-tertiary">
-                                        Connected to: <strong>{{ printer_serial }}</strong>
+                                        Connected to: <strong>{{ printer.name | e }} ({{ printer.sn }})</strong>
```
chrivers, line 150: I read up on Flask and autoescaping.

It seems that autoescaping is enabled for all templates that end in (among other extensions), `.html`, when using `render_template()`. So that covers us here, and that matches with your tests.

So we have to *not* escape, since otherwise we render some names incorrectly. For example, a printer called `Foo & Bar` would come out wrong (probably rendered as `Foo &amp; Bar`, from `Foo &amp;amp; Bar`

web/util.py
```diff
@@ -1,3 +1,5 @@
+import logging as log
+from dataclasses import asdict
```
chrivers, line 2: These imports no longer do anything. Please revert :)

chrivers (changes requested):
Almost there, but there's a couple of bugs. Fixes included :)

web/util.py
billyjbryant: ✅

static/index.html
billyjbryant: ✅

cli/config.py
billyjbryant: ✅

billyjbryant (commented):
@chrivers addressed.

chrivers (approved):
Looks good, let's get this merged :sunglasses:

PR #67, opened 2023-05-10
Upcoming/v1.0.0

just-trey:
I rebased off of @billyjbryant 's latest and made my UI adjustments. FWIW, I hate Bootstrap, but in this case, it has its place and seems like all we need for this. I will need someone to test as it actually does not work for me because I have multiple machines, but I do not expect any issues. I only modified the HTML and CSS files the rest of the changes are due to a rebase of @billyjbryant's latest


billyjbryant:
This needs to be re-based off of https://github.com/Ankermgmt/ankermake-m5-protocol/tree/upcoming/v1.0.0 now and should clean up the history quite a bit.

PR #69, opened 2023-05-11
Print Progress

billyjbryant:
### Description

This PR is an addendum to #52 and adds printer status and progress info to the Home Tab

image


chrivers:
This is very exciting! :)

There's one thing I think that detracts a lot from good work you put in, and it's the fact that all the right-hand sides of the box pairs (elapsed, remain, layer, speed, etc) are all different widths. Same goes for the two top boxes with icons.

Isn't there some way to make this a little more controlled?

I really like the idea.

thomasjpatterson:
@chrivers If I'm not mistaken that's a bootstrap thing so unless we jumped to a different underlying framework for the CSS that's the way it'll be.

billyjbryant:
> There's one thing I think that detracts a lot from good work you put in, and it's the fact that all the right-hand sides of the box pairs (elapsed, remain, layer, speed, etc) are all different widths. Same goes for the two top boxes with icons.


> @chrivers If I'm not mistaken that's a bootstrap thing so unless we jumped to a different underlying framework for the CSS that's the way it'll be.

Yes, @chrivers , @thomasjpatterson is correct, because I am using bootstrap it is just part of that and determines button size based on content. In this case I'm using a form-group with a Button as a label and a disabled input field to display the content. This presents a nice interface but I do agree the sizing of the labels might be a tad distracting. I will see what I can do to fix that or I may call upon @just-trey's expertise here to see if he can improve upon it :)

just-trey:
My gut for now tells me to leave it as is and maybe migrate away from bootstrap in a future release.

billyjbryant:
I tweaked it a bit but by doing so it is no longer responsive. 
image

billyjbryant:
> @billyjbryant  why are you doing the UI like that - why not use:
> https://getbootstrap.com/docs/5.3/forms/input-group/

I'm not quite sure what you mean because that's what I'm using... Did you mean this to be a quoted reply on a specific area of code?

static/ankersrv.js
```diff
@@ -50,12 +50,107 @@ $(function () {
...
+        return weight;
+    }
+
     /**
      * Opens a websocket connection and outputs any incoming message data to console
      */
     socket = new WebSocket("ws://" + location.host + "/ws/mqtt");
     socket.addEventListener("message", (ev) => {
-        console.log(JSON.parse(ev.data));
+        const data = JSON.parse(ev.data);
```
chrivers, line 122: We are doing a lot of this same kind of mqtt parsing in the backend, for the upcoming mainsail branch.

In the short term, I'm ok with also doing the work in the frontend, as long as it's not much bigger than this, but if we plan to parse more messages like this, we should *really* consider importing the new parsing code meant for mainsail, so we don't double the efforts.

static/img/bed-icon.svg
chrivers, line 1: What's the origin of these icons? Did you make them, or are they copied from some known source? This needs to be documented.

What's the license of the source material?

static/ankersrv.css
```diff
@@ -1,41 +1,79 @@
...
+    display: inline-block;
+    position: fixed;
+    height: 25px;
+    width: 25px;
+}
+
+.icon-hotend {
+    background-color: #fff;
+    -webkit-mask-image: url(../static/img/hotend-icon.svg);
+    mask-image: url(../static/img/hotend-icon.svg);
```
chrivers, line 62: Wait, why `../static`?

How about just `img/hotend-icon.svg`?

static/ankersrv.css
```diff
@@ -1,41 +1,79 @@
...
+}
+
+.progress {
+    height: 40px;
+}
+
+.icon {
+    background-size: cover;
+    display: inline-block;
+    position: fixed;
```
chrivers, line 54: Why are we using `position: fixed`, if we are not specifying the position?

static/index.html
```diff
@@ -131,21 +132,240 @@ <h4>{{ printer_serial }}</h4>
...
+                                        <div class="col-lg-12 px-3 rounded-3">
+                                            <div class="row mt-3">
+                                                <h4>Temperature</h4>
+                                            </div>
+                                            <div class="row mt-2">
+                                                <div class="col-6">
+                                                    <div class="input-group">
+                                                        <div class="input-group-prepend print-temp-label">
+                                                            <div class="input-group-text mx-auto" id="nozzleGroup">
+                                                                &nbsp;<i class="icon-hotend icon"></i>
```
chrivers, line 175: Since `.icon` is our own css class, can't we add some spacing attribute to it, instead of this `&nbsp;` 90's layout hack? :smile:

LazeMSS:
> > @billyjbryant  why are you doing the UI like that - why not use:
> > https://getbootstrap.com/docs/5.3/forms/input-group/
> 
> I'm not quite sure what you mean because that's what I'm using... Did you mean this to be a quoted reply on a specific area of code?

Basicaly all the input groups etc: https://github.com/billyjbryant/ankermake-m5-protocol/blob/billybryant/print_progress/static/index.html#L172

I will try and do a checkout and PR on this asap

static/ankersrv.js
```diff
@@ -50,12 +50,107 @@ $(function () {
...
+        return Math.round(temp / 100);
+    }
+
+    /**
+     * Calculate the percentage between two numbers
+     * @param {number} layer
+     * @param {number} total
+     * @returns {number} percentage
+     */
+    function getPercentage(progress) {
```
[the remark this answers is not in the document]
LazeMSS: const progress = getPercentage(data.progress);

static/ankersrv.js
```diff
@@ -50,12 +50,107 @@ $(function () {
...
+        return Math.round(((progress / 100) * 100) / 100);
+    }
+
+    /**
+     * Truncate a string to a specific length
+     * @param {string} str
+     * @param {number} maxLength
+     * @returns {string} Truncated string
+     */
+    function truncStr(str, maxLength) {
```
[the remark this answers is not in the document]
LazeMSS: And why not use .substring(0,length); and the use this class for the divs: https://getbootstrap.com/docs/5.3/helpers/text-truncation/

static/ankersrv.js
LazeMSS: Or maybe just less if states - it could be made easier to ready at least or more template friendly:
https://github.com/LazeMSS/ankermake-m5-protocol/blob/9b867cd6bac90ac5b00c25c95794d11ee9732326/static/ankersrv.js#L38

static/img/bed-icon.svg
billyjbryant: I will look up the license, they came from a royalty free svg site, I'll make sure I get the proper license in place.

LazeMSS: https://www.iconfinder.com/iconsets/3d-printing-line
license: https://creativecommons.org/licenses/by/3.0/

static/ankersrv.css
billyjbryant: This prevents the icon from tiling. If someone like @just-trey has a better way to do this, I'm open to hear it :)

static/index.html
LazeMSS: Yes basically these are the input groups I was writing about: 
https://getbootstrap.com/docs/5.3/forms/input-group/

static/ankersrv.css
thomasjpatterson: You are correct this is the easiest way to keep an image from trying to tile

LazeMSS:
In general spacing/padding/margin:
https://getbootstrap.com/docs/5.3/utilities/spacing/ 

Spinners:
there is native css based spinners for that: https://getbootstrap.com/docs/5.3/components/spinners/ - avoids using images.

Generic css :
- !important should be a final resort
- text-decoration: none -> https://getbootstrap.com/docs/5.3/utilities/text/#text-decoration
- font-size: https://getbootstrap.com/docs/5.3/utilities/text/#font-size
- font-weight: https://getbootstrap.com/docs/5.3/utilities/text/#font-weight-and-italics

Basically there is "classes" for almost anything in bootstrap - that makes is more easier to maintain and does not break on some browsers etc.

static/index.html
billyjbryant: The hack was actually to force the height of the bootstrap label, spacing wasn't a concern. Every time I took it out, no matter which bootstrap height class I used, the label was smaller than the height of the icon.

thomasjpatterson: you'd be surprised by the amount of `@nbsp;` in code to get it to move over a space.

static/ankersrv.css
LazeMSS: why not use
background-repeat: no-repeat

static/img/bed-icon.svg
billyjbryant: Thank you for the assist @LazeMSS, I'll update the PR to add proper attribution.

billyjbryant:
@LazeMSS 

> Spinners: there is native css based spinners for that: https://getbootstrap.com/docs/5.3/components/spinners/ - avoids using images.

I'm not sure what you are referring to regarding a spinner? If you are mentioning the `poster` attribute of the HTML5 `<video>` tagset, how would you suggest implementing a spinner here? 


> !important should be a final resort

I always use `!important` as a final resort. Bootstrap's CSS, despite changing the ordering in the html, is loading after the `ankerctl.css` file. In order to override certain aspects of Bootstrap, I was relegated to using `!important`, if you can find another way to do this, by all means, you are more than welcome to update it to function properly without using `!important`. 

> text-decoration: none -> https://getbootstrap.com/docs/5.3/utilities/text/#text-decoration
font-size: https://getbootstrap.com/docs/5.3/utilities/text/#font-size
font-weight: https://getbootstrap.com/docs/5.3/utilities/text/#font-weight-and-italics

I'm more than happy to shift to these, I will say; however, that the bootstrap classes for text-decoration were not functioning properly on the ONLY link that I implemented the `text-decoration:none` style attribute for.

static/ankersrv.js
```diff
@@ -50,12 +50,107 @@ $(function () {
...
+    }
+
     /**
      * Opens a websocket connection and outputs any incoming message data to console
      */
     socket = new WebSocket("ws://" + location.host + "/ws/mqtt");
     socket.addEventListener("message", (ev) => {
-        console.log(JSON.parse(ev.data));
+        const data = JSON.parse(ev.data);
+        if (data.commandType == 1001) {
```
chrivers, line 123: Since we have `.stl` files for the protocol specification, we can easily generate some a `.js` file with symbolic names for these MQTT constants.

This would make it a lot easier to follow what the decode logic is here.

I've pushed a new branch that demonstrates this. (https://github.com/Ankermgmt/ankermake-m5-protocol/blob/chrivers/basic-libflagship-js-support/static/libflagship.js)

If that works, I'll get it merged into the upcoming branch.

static/index.html
```diff
@@ -240,7 +245,10 @@ <h4>Print Progress</h4>
                                                     <div class="col-6">
                                                         <div class="input-group">
                                                             <div class="input-group-prepend">
-                                                                <div class="input-group-text" id="timeElapsed">
+                                                                <div
+                                                                    class="input-group-text print-info-label"
```
chrivers, line 249: We're adding this specific class to a number of elements here, and they all seem to be selectable by a rule such as `input-group-prepend input-group-text` or similar.

Do we really need a new css class here?

LazeMSS:
> @LazeMSS
> 
> > Spinners: there is native css based spinners for that: https://getbootstrap.com/docs/5.3/components/spinners/ - avoids using images.
> 
> I'm not sure what you are referring to regarding a spinner? If you are mentioning the `poster` attribute of the HTML5 `<video>` tagset, how would you suggest implementing a spinner here?

I would implement a div that contains the "Please wait" in text and with a css spinner. Make the video layer hidden until the video play event happens (https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play_event) then hide the loading div and show the video layer.

> 
> > !important should be a final resort
> 
> I always use `!important` as a final resort. Bootstrap's CSS, despite changing the ordering in the html, is loading after the `ankerctl.css` file. In order to override certain aspects of Bootstrap, I was relegated to using `!important`, if you can find another way to do this, by all means, you are more than welcome to update it to function properly without using `!important`.
> 
I would still say a need to use important is breaking - so if we switch to a new bootstrap theme/color (like add support for dark/normal mode) then these colors are forced instead of using the colors defined in the scheme.

> > text-decoration: none -> https://getbootstrap.com/docs/5.3/utilities/text/#text-decoration
> > font-size: https://getbootstrap.com/docs/5.3/utilities/text/#font-size
> > font-weight: https://getbootstrap.com/docs/5.3/utilities/text/#font-weight-and-italics
> 
> I'm more than happy to shift to these, I will say; however, that the bootstrap classes for text-decoration were not functioning properly on the ONLY link that I implemented the `text-decoration:none` style attribute for.

I don't even understand why the bottom link "View on github" is  btn-link class - why not just a normal link?

static/ankersrv.css
just-trey: @chrivers, this is fixed with my push. also, you will see new clock icons added are custom-built by me. What I found online was too fat.

static/ankersrv.css
just-trey: Stacik is where all the static resources are. Do we need that changed? The index and CSS are in there too.

static/index.html
just-trey: &sbsp is now gone with the latest fixes.

static/img/bed-icon.svg
chrivers: @billyjbryant Thanks. This is a MUSTFIX, since otherwise it's a violation of the upstream license.

static/index.html
chrivers: @just-trey awesome, thanks :)

static/img/load-screen.svg
chrivers, line 1: This is super cool, @just-trey. Nice work :+1:

static/img/bed-icon.svg
chrivers: @just-trey confirmed that the new icons are his original work, and so there's no upstream license to worry about.

static/ankersrv.css
chrivers: Looks great, thanks.

static/ankersrv.js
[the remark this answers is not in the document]
chrivers: I'll add a couple of commits after merging this.

static/index.html
billyjbryant: > We're adding this specific class to a number of elements here, and they all seem to be selectable by a rule such as `input-group-prepend input-group-text` or similar.
> 
> Do we really need a new css class here?

I did this so that we only effected these fields and not any other field also using the `input-group-text` class (which is used on the Upload form for the config file.

static/ankersrv.css
```diff
@@ -1,41 +1,79 @@
...
+
+.icon-bed {
+    background-color: #fff;
+    -webkit-mask-image: url(../static/img/bed-icon.svg);
+    mask-image: url(../static/img/bed-icon.svg);
+}
+
+.print-info-label {
+    width: 75px;
+    padding-left: 5px;
```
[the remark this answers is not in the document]
billyjbryant: This was a forced fix to make the labels a set width, I tried using bootstrap classes and it wasn't working properly. This addresses your original comment regarding the labels being different sizes across the elements.

static/ankersrv.css
```diff
@@ -1,41 +1,79 @@
...
+.icon-hotend {
+    background-color: #fff;
+    -webkit-mask-image: url(../static/img/hotend-icon.svg);
+    mask-image: url(../static/img/hotend-icon.svg);
+}
+
+.icon-bed {
+    background-color: #fff;
+    -webkit-mask-image: url(../static/img/bed-icon.svg);
+    mask-image: url(../static/img/bed-icon.svg);
```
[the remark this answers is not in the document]
billyjbryant: They are black SVGs, I did this here so that I could override the color without having to modify the SVG itself and found no other good way to do this.

static/ankersrv.js
```diff
@@ -50,12 +50,107 @@ $(function () {
...
+    }
+
+    /**
+     * Get the filament weight based on its length
+     * @param {number} length
+     * @returns {number} Weight
+     */
+    function getWeight(length) {
+        const diameter = 1.75;
+        const density = 1.24;
```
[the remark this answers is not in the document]
billyjbryant: This function actually needs to go away. I was originally using it to convert how much Filament was being used because the print I was doing was returning `mm` for the filament measurement but it didn't seem to be accurate so I stripped it from the visible fields.

static/ankersrv.css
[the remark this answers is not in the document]
chrivers: Ah, that makes sense. Thanks for the explanation :)

static/ankersrv.css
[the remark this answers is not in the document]
chrivers: Ah, it seems bootstrap has some weird quirks with the way sizing interacts with this. And with @just-trey's changes on top, I'd consider this resolved now :)

chrivers (approved):
Looks like we've addressed all conversations, so let's get this thing merged :)

Issue #47, opened 2023-04-11
Cert error when executing mqtt command but not from the ankerctl directory


dmbaio:
I get the following error if I run the python file with an mqtt option, but only if I'm not CD'd into that directory.

![image](https://user-images.githubusercontent.com/59879282/231025904-25fc9ab6-0f51-48f0-9267-39ce8e514044.png)

chrivers:
Thanks for reporting this - I'll see if I can get it fixed soon.

We need to add some code to the python script, to locate the certificate file, regardless of the current working directory

thomasjpatterson:
Thanks @dmbaio, we've been furiously working behind the scenes and one of the features that will be upcoming is the ability to use a web interface to select your login.json file for use so you can have it anywhere. Once #59 is complete you'll have this capability.

billyjbryant:
This is addressed by #71 and will be released with v1.0.0

thomasjpatterson:
Ready to Ship.

PR #71, opened 2023-05-12
Fixing issue with MQTT not finding SSL Cert

billyjbryant:
### Description

Currently, references to the ssl cert for MQTT are relative and the cert is stored in a directory called `examples` that does not exist when the docker container is built. 

This PR fixes this by moving the SSL cert to the path `./ssl/` which is copied to the Docker container and updates references to it to use Absolute path, this way you can start the service from any directory without issue. 

See image for errors that will be fixed by this pr : 

image


austinrdennis:
Tested on Windows 10 Pro (64 Bit) using Docker via WSL2. Bug squashed, recommend merge.

billyjbryant:
This fixes #47

billyjbryant:
Other issues I corrected:

* https://github.com/Ankermgmt/ankermake-m5-protocol/issues/47
* Incorrectly setting cert validation to `False` by using `not` when resolving the value of the `insecure` arg
* Provides mechanism for obtaining the true absolute path by including `config.ROOT_DIR` and using `os.path.join` to join it with a relative file path.

spuder:
There is a whole lot going on in this PR.

I think it might be best to split into smaller easier to grok PRs so that each feature/enhancement can be reviewed individually. 


I can comment on the docker and .gitattributes sections, but not so much on the python sections.

docker-compose.yaml
```diff
@@ -1,26 +1,26 @@
...
-    environment:
-      - FLASK_PORT=4470
-      - FLASK_HOST=0.0.0.0
-    volumes:
-      - ankerctl_vol:/root/.config/ankerctl
-    ports:
-      - 127.0.0.1:4470:4470
-    entrypoint: "/app/ankerctl.py"
-    command: ["webserver", "run"]
+    ankerctl:
```
spuder, line 9: Changing the line indentations from 2 spaces to 4 spaces makes it kind of hard to see what changed.

it looks like working_dir was added?

could we keep the indenting the same and handle a linting refactor in a separate PR? (Ideally automated to accommodate different editor defaults)

billyjbryant: See #72

billyjbryant:
> I think it might be best to split into smaller easier-to-grok PRs so that each feature/enhancement can be reviewed individually.

@spuder I moved all changes for Docker or related to building the docker image to #72 :)

libflagship/mqttapi.py
```diff
@@ -66,12 +71,13 @@ def on_message(self, client, userdata, msg, pkt, tail):
...
 
     @classmethod
-    def login(cls, printersn, username, password, key, ca_certs="ankermake-mqtt.crt", verify=True):
+    def login(cls, printersn, username, password, key, ca_certs=cert_path, verify=True):
         client = mqtt.Client()
 
         if verify:
             client.tls_set(ca_certs=ca_certs)
         else:
+            log.warning('Not verifying certificates is insecure')
```
spuder, line 80: Could this be rephrased to be a positive statement instead of a double negative? 

By not verifying the certificate is insecure, is that the same as 'skipping certificate verification' ?

cli/mqtt.py
```diff
@@ -22,8 +26,8 @@ def mqtt_open(config, insecure):
             acct.mqtt_username,
             acct.mqtt_password,
             printer.mqtt_key,
-            ca_certs="examples/ankermake-mqtt.crt",
-            verify=not insecure,
+            ca_certs=cert_path,
+            verify=insecure,
```
spuder, line 30: Could you walk me through what this line is doing? 

`mqtt_open()` is called twice
1. cli
2. web/service

When called as part of web/service, then the `insecure` is set to `True` (line 14)
When called as part of the cli it is set to `env.insecure` which is defaulting to `env.insecure = insecure` on line 79 of the main() function.

I'll probably defer this to @chrivers on this one because I'm not sure why the 2 levels of indirection for this boolean.

billyjbryant: Ok, so I found the error and discovered my err. I assumed we were setting `insecure` incorrectly here but it was actually in the webserver that we were doing so. I have reverted this change here and set up the webserver to allow passing the insecure flag through to the app configuration for flask.

libflagship/mqttapi.py
billyjbryant: I have made this more verbose and cleared up the language to provide suggested use case.

billyjbryant:
I also added docstrings to a few of the modules in an effort to slowly expand the quality of internal documentation.

chrivers:
> There is a whole lot going on in this PR.
> 
> I think it might be best to split into smaller easier-to-grok PRs so that each feature/enhancement can be reviewed individually.
> 
> I can comment on the docker and .gitattributes sections, but not so much on the python sections.

Absolutely agree.

@billyjbryant I really appreciate all the work you put into this, but these "mixed bag" PRs are so much harder to review, since they change a whole bunch of unrelated things.

It would be much appreciated if you made smaller, more focused PRs. (for instance, the docstring improvements could go in their own PR)

config/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+Imports: - os.path: Provides functions to interact with file paths.
+
+Variables: - ROOT_DIR: A string containing the absolute path of the parent directory of the file this code is in.
+
+Usage: - Import this module into your script, then access the ROOT_DIR variable \
+to get the absolute path of the parent directory for your project. """
+from os import path
+
+ROOT_DIR = path.realpath(path.join(path.dirname(__file__), '..'))
```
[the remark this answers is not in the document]
chrivers: Oh, and as we discussed, I consider this a must fix :)

examples/mqtt-connect.py
```diff
@@ -122,7 +122,7 @@ def main():
 
     # get up and running
     try:
-        client = AnkerMQTTClient.login(printer_sn, mqtt_username, mqtt_password, mqtt_key, verify=not args.insecure)
+        client = AnkerMQTTClient.login(printer_sn, mqtt_username, mqtt_password, mqtt_key, verify=args.insecure)
```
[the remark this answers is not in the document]
billyjbryant: I reverted this change, i had not intended to do this (I mistakenly used a find and replace)

cli/mqtt.py
```diff
@@ -12,6 +14,8 @@
 
 
 def mqtt_open(config, insecure):
+    cert_path = path.join(ROOT_DIR, "ssl/ankermake-mqtt.crt")
```
chrivers, line 17: No need to import `os.path`:

```python
cert_path = ROOT_DIR / "ssl/ankermake-mqtt.crt"
```

(that's the new Pathlib.Path syntax)

cli/mqtt.py
```diff
@@ -22,7 +26,7 @@ def mqtt_open(config, insecure):
             acct.mqtt_username,
             acct.mqtt_password,
             printer.mqtt_key,
-            ca_certs="examples/ankermake-mqtt.crt",
+            ca_certs=cert_path,
```
chrivers, line 29: We can just write `ROOT_DIR / "ssl/ankermake-mqtt.crt"` here, without need for `cert_path`

libflagship/mqttapi.py
```diff
@@ -1,14 +1,19 @@
 import paho.mqtt.client as mqtt
 import paho.mqtt
+from os import path
 import logging as log
 import ssl
 import json
 import uuid
 from datetime import datetime, timedelta
 
+from config import ROOT_DIR
```
chrivers, line 10: This is a problem - `libflagship` does not include anything except its own modules, and dependencies from `requirements.txt`, but now also this from a "foreign" directory (beyond top level of `libflagship/`).

This makes it impossible to refactor `libflagship` into a independent module, which is something we might want later.

Fortunately, the solution is super simple, see below :)

libflagship/mqttapi.py
```diff
@@ -1,14 +1,19 @@
...
 from libflagship.mqtt import MqttMsg, MqttPktType
 
 
+cert_path = path.join(ROOT_DIR, "ssl/ankermake-mqtt.crt")
```
chrivers, line 14: Remove this, see below.

libflagship/mqttapi.py
```diff
@@ -66,13 +71,16 @@ def on_message(self, client, userdata, msg, pkt, tail):
...
         client = mqtt.Client()
 
         if verify:
             client.tls_set(ca_certs=ca_certs)
         else:
-            client.tls_set(ca_certs=ca_certs, cert_reqs=ssl.CERT_NONE)
+            log.warning('[Not Verifying Certificate]')
+            log.warning('This is insecure and should not be used in production environments.')
+            log.warning('It is recommended to run without "-k/--insecure".')
+            client.tls_set(ca_certs=None, cert_reqs=ssl.CERT_NONE)
```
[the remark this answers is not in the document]
chrivers: Let's keep libflagship simple, by defaulting `ca_certs=None`, and not having a real path default here. We are calculating it where needed anyway, so that'd be double the work.

This makes `libflagship` just a tool to be used, and keeps the glue logic in the cli part. Also, it's simpler, we just remove some lines :)

(must fix)

chrivers (changes requested):
Almost there - a couple of minor fixes and simplifications, and we are good to go :+1:

libflagship/mqttapi.py
billyjbryant: ✅

libflagship/mqttapi.py
billyjbryant: ✅

cli/mqtt.py
billyjbryant: ✅

cli/mqtt.py
billyjbryant: ✅

billyjbryant (commented):
Requested changes addressed. @chrivers please re-review.

chrivers (approved):
Looks great! Let's get this merged :sunglasses:

PR #72, opened 2023-05-13
Fixing issues with building docker image

billyjbryant:
### Description
* Fixing issues with building docker image on windows
* Removing the external volume indication since config can be uploaded from web now
* Set the working_dir in Dockercompose to /app

Prettier buried the lead on `docker-compose.yaml`, actual changes were: 
```diff
services:
  ankerctl:
    image: ankerctl/ankerctl:latest
    container_name: ankerctl
    restart: unless-stopped
    build: .
    environment:
      - FLASK_PORT=4470
      - FLASK_HOST=0.0.0.0
    volumes:
      - ankerctl_vol:/root/.config/ankerctl
    ports:
      - 127.0.0.1:4470:4470
+   working_dir: /app
    entrypoint: "/app/ankerctl.py"
    command: ["webserver", "run"]

volumes:
  ankerctl_vol:
-   external: true
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,4 +7,4 @@
 *.md    text
 *.txt   text
 *.tpl   text
-*.py    text diff=python
+*.py    text diff=python eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN mkdir -p /root/.config/
 # Copy the script and libraries
 COPY ankerctl.py /app/
 COPY web /app/web/
+COPY ssl /app/ssl/
 COPY static /app/static/
 COPY libflagship /app/libflagship/
 COPY cli /app/cli/
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,6 @@ RUN mkdir -p /root/.config/
 # Copy the script and libraries
 COPY ankerctl.py /app/
 COPY web /app/web/
-COPY ssl /app/ssl/
 COPY static /app/static/
 COPY libflagship /app/libflagship/
 COPY cli /app/cli/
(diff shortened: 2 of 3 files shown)


spuder:
Excellent work, thank you for fixing this. 

I now see that there is actually a new issue introduced by splitting the PR because there is no `ssl` directory in main.

billyjbryant:
ahh, you are so right, let me move that over to the other PR :D

spuder (approved):

.gitattributes
```diff
@@ -7,4 +7,4 @@
 *.md    text
 *.txt   text
 *.tpl   text
-*.py    text diff=python
+*.py    text diff=python eol=lf
```
chrivers, line 10: Okay, explain this please.

The whole reason we added `.gitattributes` in the first place, was to provide windows users with CRLF conversion. What are we trying to do here? Why not just remove this file altogether?

I'm not trying to defend CRLF (it is an abhorrent abomination), but what are we trying to solve?

If Windows users generally have tools that can deal with LF line endings, why try to do CRLF conversion for some files?

And if they don't, why can we live with skipping this conversion for .py files?

I'm missing some consistency here guys...

billyjbryant: The issue we were looking to correct here is that if you are using Docker Desktop on Windows, the docker build runs on a linux based WSL container not your local machine. Git Desktop for Windows does auto-conversion to CRLF on all files, which causes `docker build .` to produce errors on launching the container. This was witnessed by @thomasjpatterson and @austinrdennis while trying to test the docker functions to build documentation. I am happy to roll this change out; however, we would need to potentially document how to address such errors if a user incurs them during the build of the docker image (should they choose to build their own rather than using the autobuild version).

PR #75, opened 2023-05-14
V1.0.0 documentation mac install

thomasjpatterson:
I added the Mac instructions for the Docker install. Thankfully even though there are two types of hardware the install process is the same except for the Rosetta 2 piece I highlighted in bullet 2. Please review and if you don't see any issues lets get it into your main documentation branch.

### DIFF
--- a/documentation/install-from-docker.md
+++ b/documentation/install-from-docker.md
@@ -45,7 +45,16 @@
 
 1. Install the [AnkerMake slicer](https://www.ankermake.com/software). Make sure you open it and login via the “Account” dropdown in the top toolbar.
    	**NOTE:** The slicer app does not need to be open for the rest of these steps.
-2. 
+
+2. Install Docker Desktop that matches your CPU by following [this guide](https://docs.docker.com/desktop/install/mac-install/) (supplemental steps are provided below).
+
+    It is strongly recommended that if your are using a Mac with Apple Silicon to install Rosetta 2. It is no longer a requirement since Docker 4.3.0 however for the best experience it should be installed. The easiest way to install it is via your preferred choice of Terminal and running this command:
+      
+      `softwareupdate --install-rosetta`
+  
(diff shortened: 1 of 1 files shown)


documentation/install-from-docker.md
```diff
@@ -45,10 +45,28 @@
...
+      
+      `softwareupdate --install-rosetta`
+  
+     - [ ] **[<u>Install Docker Desktop</u>]** Download and install Docker Desktop using the linked docker docs guide. Once the install process is complete, open Docker Desktop and accept the EULA. You can close Docker Desktop for the remaining steps.
+
+3. Download the latest version of ankerctl from the GitHub releases page and extract the contents wherever you would like to run the program from.
+
+4. The next steps will take place in your preferred choice of Terminal.
+
+   4a. Navigate to the ankerctl directory (i.e.: cd \Users\username\Downloads\ankerctl if you unzipped it in your Downloads folder)
```
spuder, line 61: Minor suggestion would be to use a valid path `\Users\username` won't actually work but `$HOME/Downloads` or `~/Downloads` will 

```suggestion
   4a. Navigate to the ankerctl directory (i.e.: `cd ~/Downloads/ankerctl` if you unzipped it in your Downloads folder)
```

documentation/install-from-docker.md
```diff
@@ -45,10 +45,28 @@
...
+   4b. Run the following command to build the docker container(s) and volume(s) needed by ankerctl.
+
+      `docker compose build`
```
spuder, line 64: I think we could simplify this for the following reasons

1. Users should no longer need to build images since they are being built and hosted plubically. 
2. docker compose build _should_ happen automatically if docker detects that the image isn't available publically and a Dockerfile is available in the current working directory. 


Instead we could replace this with just

```
docker compose up
```

The rest will be handled automatically.

documentation/install-from-docker.md
thomasjpatterson: Thanks for catching this, I was doing this before fully caffeinated.

documentation/install-from-docker.md
thomasjpatterson: Changed to docker compose up

austinrdennis (approved):
Nice one, you and @spuder are all over it! 👍🏻

PR #80, opened 2023-05-16
chrivers/basic-libflagship-js-support

chrivers:
I implemented a new js template for the MqttMsgType, allowing us to use symbolic constants from `ankersrv.js`. This is the support only, so `ankersrv.js` has not been changed.


chrivers:
@LazeMSS @just-trey @billyjbryant Does this seem useful for js mqtt parsing?

It's trivial to merge, but there's no reason to merge it, if it isn't wanted.

just-trey:
If we ultimately want to get to mainsail I'm wondering if it's overkill.

just-trey:
@billyjbryant https://docs.mainsail.xyz/

thomasjpatterson:
Different thought here; I say keep it and bring it in. It would allow for someone else to be able to make a frontend or couple with other products. The work is already done, might as well just stick in there for someone else to expand on if they so choose.

chrivers:
I tend to agree - but in any case, we have the branch on github now, so we can always merge it when we want. I think I might make a PR to use the symbolic names in the parser.

chrivers:
I'm merging this, since it gives us some new options, and has almost no impact on size or complexity.

PR #88, opened 2023-05-19
chrivers/webui-simplification

chrivers:


static/base.html
```diff
@@ -0,0 +1,86 @@
+{% import "macro.html" as macro -%}
+<!DOCTYPE html>
+<html data-bs-theme="dark" lang="en">
+    <head>
+        <meta charset="utf-8" />
+        <meta content="width=device-width, initial-scale=1" name="viewport" />
+        <title>{% block title %}{% endblock title %}</title>
+        <link rel="icon" href="{{ url_for('static', filename='img/logo.svg') }}">
+        {{ macro.static_css("vendor/bootstrap-5.3.0-alpha3-dist/css/bootstrap.min.css") }}
```
billyjbryant, line 9: Nice Touch! Makes it a lot easier to read.

static/tabs/setup.html
```diff
@@ -0,0 +1,85 @@
...
+                    action="{{ url_for('app_api_ankerctl_config_upload') }}"
+                    method="POST"
+                    enctype="multipart/form-data"
+                    class="mb-3"
+                    id="config-upload-form"
+                >
+                    <div class="card">
+                        <div class="card-header">
+                            <label for="loginFile" class="fs-4">
+                                Upload AnkerMake Login File
```
billyjbryant, line 20: We should probably wrap this in `<h4>` or remove the `<h4>` wrapper around the other section heads so that the font-types are the same. 

image

static/ankersrv.js
```diff
@@ -105,7 +105,7 @@ $(function () {
         const data = JSON.parse(ev.data);
         if (data.commandType == 1001) {
             // Returns Print Details
-            $("#print-details").attr("style", "display: block;");
+            $("#print-details").removeClass("d-none");
```
billyjbryant, line 108: I was thinking about this, should we show this block of details all the time? There is no event that will hide it once it has been shown, outside of a refresh.

billyjbryant (changes requested):
Looks good, a few other style fixes while we are at it :)

static/ankersrv.js
chrivers: Yeah, I was thinking the same thing while refreshing this. We don't have any good logic for keeping track of when to show this, and the user might even want to see the stats of the last print job after it's done.

I'll just make this always-on :+1: :white_check_mark:

static/tabs/setup.html
chrivers: Good point - I actively removed the `<h4>` here, since neither `<label>` or `<h4>` is allowed to nest inside the other.

I cleaned this up to use `.fs-4` in all 3 places, and removed all `<h4>` tags on this page.

:white_check_mark:

chrivers:
Changes since last review:

 - Remove `.print-details` wrapper, and make print details always shown
 - Combine all 4 video control buttons into one
 - Move "Connected to" headline into a new "Connection" card in the right column. In the future, we can add more detailed connection information here (mqtt, pppp, steaming bandwidth, wifi strength updates, etc)
 - Normalized some mixed use of `<h4>`, `.h6`, `.fs-6`, etc.

thomasjpatterson:
When restarting ankerctl services it hangs on the spinner and `MqttQueue: Worker Stopped` until I manually refresh the screen. Mac M2 Chrome

thomasjpatterson:
[manualrefresh.log](https://github.com/Ankermgmt/ankermake-m5-protocol/files/11517026/manualrefresh.log)
Line 5536

thomasjpatterson:
Can confirm, new commit resolves the issue.

static/tabs/home.html
```diff
@@ -0,0 +1,223 @@
...
+                                    >
+                                        00:00:00
+                                    </span>
+                                </div>
+                            </div>
+                        </div>
+                        <div class="row mb-3 g-3">
+                            <div class="col-12">
+                                <div class="progress position-relative" id="printProgress">
+                                    <div
```
billyjbryant, line 169: With the change of the color for the progress bar, we should probably change the font color so that there is greater contrast. 

image

static/tabs/instructions.html
```diff
@@ -0,0 +1,45 @@
+<div class="tab-pane fade" id="instructions" role="tabpanel" aria-labelledby="instructions-tab">
+    <div class="container mb-4">
+        <div class="row g-3">
+            <div class="col-lg-6">
+                <h2>Connecting PrusaSlicer/SuperSlicer</h2>
+                <ol class="list-group list-group-numbered mb-3">
+                    <li class="list-group-item">Go to "Printer settings" tab</li>
+                    <li class="list-group-item">Click "Gear"⚙️ (Add/Edit Physical Printer)</li>
```
billyjbryant, line 8: Since we have shifted to using BS icons for everything else, maybe we should use one here for the gear.

billyjbryant (commented):
Few comments.

static/tabs/home.html
thomasjpatterson: good call on that, that almost reminds me of a colorblind challenge.

static/tabs/instructions.html
chrivers: Done :white_check_mark:

static/tabs/home.html
chrivers: Fixed :white_check_mark:

static/tabs/instructions.html
```diff
@@ -0,0 +1,48 @@
+<div class="tab-pane fade" id="instructions" role="tabpanel" aria-labelledby="instructions-tab">
+    <div class="container mb-4">
+        <div class="row g-3">
+            <div class="col-lg-6">
+                <h2>Connecting PrusaSlicer/SuperSlicer</h2>
```
just-trey, line 5: Semantically speaking this should be an h1. We can apply h2 styling via boostrap

just-trey (commented):
@chrivers one minor suggestion if you want to make it. This is awesome

static/tabs/instructions.html
chrivers: Oh, good catch. I'll fix!

chrivers: Done :white_check_mark:

PR #89, opened 2023-05-19
Import login.json automatically in docker compose

spuder:
Solves the problem that was discussed in the Discord which I think will make everyone happy 😄 

Problem 1: Improve user experience so they only need to run 1 `docker compose` command (instead of 2) 
Problem 2: Attempt to automatically discover login.json on mac/linux machines, and if found import them. (Eliminating the need to manually import the config file in the gui. Users do not need to memorize `cmd+shift+g` or permenantly alter their ~/Library visibility settings)

## How it works

- Adds 3 new docker compose steps that ankerctl service depends on
( ankerctl_seed_mac, ankerctl_seed_linux, ankerctl_import)
- Attempt to bind mount to hard coded mac/linux paths for login.json
- Copies login.json to /tmp/login.json in the docker container
- runs `ankerctl config import` to copy the config to the docker volume
- All of these commands run with `|| true` to ignore any failures. Containers run with `restart: "no"` to happily continue if the import fails (falling back to the default behavior of manually importing config)

## Known issues / Compromises

This doesn't solve the issue where the config file has to be re-imported every time the container starts, however by automating the import as part of the startup, that is effectively a non-issue.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,26 +1,55 @@
-version: '3.9.1'
+version: "3.9.1"
 
-# You must manually run `ankerctl config import` atleast once
-# docker run \
-# -v ankerctl_vol:/root/.config/ankerctl \
-# -v "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json" \
-# ankerctl config import /tmp/login.json
 services:
-  ankerctl:
-    image: ankerctl/ankerctl:latest
-    container_name: ankerctl
-    restart: unless-stopped
-    build: .
-    environment:
-      - FLASK_PORT=4470
-      - FLASK_HOST=0.0.0.0
-    volumes:
-      - ankerctl_vol:/root/.config/ankerctl
-    ports:
-      - 127.0.0.1:4470:4470
-    entrypoint: "/app/ankerctl.py"
-    command: ["webserver", "run"]
+    ankerctl_seed_mac:
+        image: busybox
+        container_name: ankerctl_seed_mac
+        restart: "no"
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+            - "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json"
+        command: sh -c "cp /tmp/login.json /root/.config/ankerctl/login.json || true"
+    ankerctl_seed_linux:
+        image: busybox
+        container_name: ankerctl_seed_linux
+        restart: "no"
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+            # - "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json" #TODO: Change to linux path
+        command: sh -c "cp /tmp/login.json /root/.config/ankerctl/login.json || true"
+    ankerctl_import:
+        image: ankerctl/ankerctl:latest
+        container_name: ankerctl_import
+        restart: "no"
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+            - "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json"
+        entrypoint: "/app/ankerctl.py"
+        command: ["config", "import", "/tmp/login.json"]
+        depends_on:
+            - ankerctl_seed_mac
+            - ankerctl_seed_linux
+        
+    ankerctl:
+        image: ankerctl/ankerctl:latest
+        container_name: ankerctl
+        restart: unless-stopped
+        build: .
+        environment:
+            - FLASK_PORT=4470
+            - FLASK_HOST=0.0.0.0
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+        ports:
+            - 127.0.0.1:4470:4470
+        working_dir: /app
+        entrypoint: "/app/ankerctl.py"
+        command: ["webserver", "run"]
+        depends_on:
+            - ankerctl_seed_mac
+            - ankerctl_seed_linux
+            - ankerctl_import
 
 volumes:
   ankerctl_vol:
-    external: true
+    driver: internal


docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
...
-    command: ["webserver", "run"]
+    ankerctl_seed_mac:
+        image: busybox
+        container_name: ankerctl_seed_mac
+        restart: "no"
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+            - "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json"
+        command: sh -c "cp /tmp/login.json /root/.config/ankerctl/login.json || true"
+    ankerctl_seed_linux:
```
billyjbryant, line 12: There is no version of AnkerMake for Linux just yet, and if there is in the future the path will look more like `$HOME/.local/share/<AnkerMakeSpecificPath>`

docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
...
+        # image: ghcr.io/ankermgmt/ankermake-m5-protocol:latest
+        image: ankerctl/ankerctl:latest
+        container_name: ankerctl_import
+        restart: "no"
+        build: .
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+        entrypoint: "/app/ankerctl.py"
+        command: ["config", "import", "/root/.config/ankerctl/login.json"]
+        depends_on:
```
billyjbryant, line 31: No `ankerseed` for Windows? Bind volumes work in Windows too.

billyjbryant (changes requested):
There isn't an official version of AnkerMake for Linux at the moment, the only way to circumvent this is with Wine. I think we should replace `ankerseed_linux` with `ankerseed_win` and let Linux users fend for themselves with the browser-based upload.

docker-compose.yaml
thomasjpatterson: INFORMATIONAL: Not for v1.0.0 ship but for v.1.1 we can look at their alpha build of the new slicer since that does have a linux build.

docker-compose.yaml
thomasjpatterson: Isn't this just a naming thing though? It's called ankerctl_import and is still importing the login.json file right?

thomasjpatterson:
Not working on Mac M2 
CleanShot 2023-05-19 at 07 55 10
CleanShot 2023-05-19 at 07 55 39

docker-compose.yaml
thomasjpatterson: Ok after futzing with this for about an hour I have it working this way:

`volumes:
    - $HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json 
    #command: sh -c "cp /tmp/login.json /root/.config/ankerctl/login.json || true"`
  NO command argument

chrivers:
Should this target `upcoming/v1.0.0`?

docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
...
+            - "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json"
+        command: sh -c "cp /tmp/login.json /root/.config/ankerctl/login.json || true"
+    ankerctl_seed_linux:
+        image: busybox
+        container_name: ankerctl_seed_linux
+        restart: "no"
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+            # TODO: verify this is the correct linux path
+            - "$HOME/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json"
```
chrivers, line 19: This would be our best bet, for Linux users running ankermake through wine with default settings:

```sh
$HOME/.wine/drive_c/users/$USER/AppData/Local/AnkerMake/AnkerMake_64bit_fp/login.json
```

docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
...
+            - ankerctl_vol:/root/.config/ankerctl
+        entrypoint: "/app/ankerctl.py"
+        command: ["config", "import", "/root/.config/ankerctl/login.json"]
+        depends_on:
+            - ankerctl_seed_mac
+            - ankerctl_seed_linux
+        
+    ankerctl:
+        # image: ghcr.io/ankermgmt/ankermake-m5-protocol:latest
+        image: ankerctl/ankerctl:latest
```
chrivers, line 37: Can you explain how `ankerctl/ankerctl:latest` works? I'm not quite familiar with that syntax.

Suppose we make sure `ghcr.io/ankermgmt/ankermake-m5-protocol:latest` always has a fresh image available. What are the pros/cons of using that image here, vs pointing to `ankerctl/ankerctl`, which (I imagine) points to the locally-built image?

spuder:
> Should this target `upcoming/v1.0.0`?

Its close, but I would say don't delay the release for it.

docker-compose.yaml
spuder: 🤔 Hmm, that is interesting. If the command argument is commented out then the `cp` won't work. 

I suspect what happened is another previous attempt copied the login.json to the volume. 

If you purge your docker volumes, and then try again I'm fairly confident that this will not work with that line commented out. 

```
docker system prune
docker volume rm ankermake-m5-protocol_ankerctl_vol
docker compose up
```

docker-compose.yaml
spuder: Good point, lets scrap linux support, and add windows support. Do you know the path on windows?

docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
...
+    ankerctl_import:
+        # image: ghcr.io/ankermgmt/ankermake-m5-protocol:latest
+        image: ankerctl/ankerctl:latest
+        container_name: ankerctl_import
+        restart: "no"
+        build: .
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+        entrypoint: "/app/ankerctl.py"
+        command: ["config", "import", "/root/.config/ankerctl/login.json"]
```
chrivers, line 30: Wouldn't it be simpler to put `login.json` in (say) `/root/login.json`?

In fact, this location clashes with an upcoming feature, where it will be possible to select other "profiles" than `default.json`. (profiles are "complete config imports" - thank of them like browser profiles).

Also, can't we stack `ankerctl_seed_mac` and `ankerctl_seed_linux`, by just trying several `cp ... || true` with different paths? We could try all well-known paths, and if one of them works, we can import it. How does that sound?

I realize now that `/root` isn't part of the volume, but since we import the config, I suppose that's actually a feature, since the copy of `login.json` will then be ephemeral, while the imported `default.json` will be kept. Seems good?

docker-compose.yaml
chrivers: The way I understand it, the `ankerctl_seed_*` containers try to copy-in a `login.json` file from the host system, and `ankerctl_import` runs the import action.

@spuder could we make `ankerctl_import` responsible for all the `cp ... || true` attempts, and also `ankerctl.py config import`? That would make this a simple two-step thing.

docker-compose.yaml
spuder: Eventually `ghcr.io/ankermgmt/ankermake-m5-protocol:latest` will be preferred, however for testing it won't work since I need the latest docker container available on my machine. 

We can switch ankerctl/ankerctl to ghcr once we have the updated web gui published. 
I can actually change this in git, and just override it temporarily while testing.

docker-compose.yaml
billyjbryant: > Good point, lets scrap linux support, and add windows support. Do you know the path on windows?

`%LOCALAPPDATA%\Ankermake\AnkerMake_64bit_fp\login.json`

billyjbryant: @spuder See above

billyjbryant: Or `c:\Users\$USER\AppData\Local\Ankermake\AnkerMake_64bit_fp\login.json`

docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
...
-      - 127.0.0.1:4470:4470
-    entrypoint: "/app/ankerctl.py"
-    command: ["webserver", "run"]
+    ankerctl_seed_mac:
+        image: busybox
+        container_name: ankerctl_seed_mac
+        restart: "no"
+        volumes:
+            - ankerctl_vol:/root/.config/ankerctl
+            - "$HOME/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json:/tmp/login.json"
```
billyjbryant, line 10: This will fail on mac because of the space, you need to properly escape the space between `Application` and `Support`.

docker-compose.yaml
billyjbryant: I've tried a few local edits and can't find a way of getting thsi to work for Windows using Docker Desktop on WSL.

chrivers:
After trying numerous techniques, it's clear that this is not quite going to be ready for v1. I've put this on the v1.1 milestone, to give us some time to figure out if this is going to work well.

In the meantime, I'm going to implement a small helper script.

PR #102, opened 2023-05-22
chrivers/packaging for v1

chrivers:
This pr fixes the networking settings for docker-compose (`network_mode: host`), which is required for pppp communications.

Also, the stop signal has been specified, and duplication between Dockerfile and docker-compose has been limited.


chrivers:
@spuder I'm particular interested in your review, if you have time :)

You have more experience with docker and -compose. Does anything seem out of place?

docker-compose.yaml
```diff
@@ -11,16 +9,19 @@ services:
...
+        # bind to localhost by default
         environment:
+            - FLASK_HOST=127.0.0.1
             - FLASK_PORT=4470
-            - FLASK_HOST=0.0.0.0
         volumes:
             - ankerctl_vol:/root/.config/ankerctl
         ports:
             - 127.0.0.1:4470:4470
-        working_dir: /app
```
spuder, line 21: Just a comment: 
currently `working_dir` is not defined in either the Dockerfile or the docker-compose. This isn't a blocker, but might cause some confusion down the line if the entrypoint or other script gets added. 

While definitely not a "best practice", I consider explicitly defining the working_dir to be a good idea.

docker-compose.yaml
```diff
@@ -11,16 +9,19 @@ services:
         container_name: ankerctl
         restart: unless-stopped
         build: .
+
+        # host-mode networking is required for pppp communication with the
+        # printer, since it is an asymmetrical udp protocol.
+        network_mode: host
```
spuder, line 15: I'm finding that on mac, that network_mode: host does not work. The connection is refused. 

```
# Steps to purge all volumes and containers
docker system prune -f
docker volume rm ankermake-m5-protocol_ankerctl_vol
docker rmi $(docker image list -q ankerctl/ankerctl)
```

I've also tried setting the more permissive (but insecure) seeings

`FLASK_HOST=0.0.0.0`
```
ports:
  - 4470:4470
```

Sadly docker desktop for mac is quite limited, yet it is the most popular docker engine (rancher being probably the second most pouplar)

From the docker documentation

> The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.

https://docs.docker.com/network/host/

## Workaround

On mac, the following works instead of `network_mode: host`, however I have no idea if that will work on linux. 

```
    extra_hosts:
      - "host.docker.internal:host-gateway"
      ```

File: docker-import.sh [hunk 3]
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
```
spuder, line 1: Might want to clarify in a comment or documentation if this script is intended to be used by humans, or if it is intended to be consumed as an entrypoint script. 



```
** No /Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json credentials detected **
** No /Library/Application Support/Ankermake/AnkerMake_64bit_fp/login.json credentials detected **
** No //Users/spencer.owen/.wine/drive_c/users/spencer.owen/AppData/Local/AnkerMake/AnkerMake_64bit_fp/login.json credentials detected **
** No //Users/spencer.owen/.wine/drive_c/users/spencer.owen/AppData/Local/Ankermake/AnkerMake_64bit_fp/login.json credentials detected **
** Importing /Users/spencer.owen/Library/Application Support/AnkerMake/AnkerMake_64bit_fp/login.json credentials **
[*] Loading cache..
[*] Initializing API..
[*] Requesting profile data..
[!] Config import failed: API request failed: 401 Unauthorized (auth token might be expired: make sure Ankermake Slicer can connect, then try again)
[*] ServiceManager: Shutdown complete
Configuration import failed :(
** Importing /Users/spencer.owen/Library/Application Support/Ankermake/AnkerMake_64bit_fp/login.json credentials **
[*] Loading cache..
[*] Initializing API..
[*] Requesting profile data..
[!] Config import failed: API request failed: 401 Unauthorized (auth token might be expired: make sure Ankermake Slicer can connect, then try again)
[*] ServiceManager: Shutdown complete
Configuration import failed :(
** No /Users/spencer.owen//Users/spencer.owen/.wine/drive_c/users/spencer.owen/AppData/Local/AnkerMake/AnkerMake_64bit_fp/login.json credentials detected **
** No /Users/spencer.owen//Users/spencer.owen/.wine/drive_c/users/spencer.owen/AppData/Local/Ankermake/AnkerMake_64bit_fp/login.json credentials detected **
```

Are users expected to run these steps (this is what I had to do to get it working)
1. `docker compose up`
2. run script
3. `docker compose restart`

spuder (approved):

docker-compose.yaml
chrivers: We have this in the non-build env in `Dockerfile`:

```
# Set the working directory to /app
WORKDIR /app
```

Does that cover it?

spuder: Oh yes, thats great.

docker-compose.yaml
chrivers: Does this work with video streaming?

Almost any network setup will work with the http traffic for flask, but the problem is the crazy udp traffic ankerctl has to send to the printer.

So unless we have verified with video streaming (or, uploading a print job), we don't know if it works.

docker-import.sh
chrivers: Ah, hm, good point.

Through the web ui, we can click the "reload services" button, but yeah, I'll incorporate a `docker restart` on the container at the end of the script.

docker-compose.yaml
spuder: My printer is torn apart at the moment so I'm not positive. 

Though after more research, it looks like the `extra_hosts` workaround is docker-desktop (mac) specific, so I don't think we should go with it. 

:shakes_fist: For being the standard for making code portable, docker sure isn't very portable is it? 

This is the closest workaround for linux (but doesn't work for mac) :shakes_fist: :shakes_fist:

```
    ports:
      - 127.0.0.1:4470:4470
    environment:
      - HOST_IP=<host-ip-address>
      ```

chrivers: Anything that relates to port 4470 is only for the http interface. UDP runs on port 32108, plus any number of randomly-opened ports (and that's part of the problem).

As far as I can tell, *any* docker network mode that involves NAT is going to break ankerctl's chance of talking to the printer.

So far, we have verified `host` and `macvlan` as working, and `default` (NAT-based) as broken. The rest are unknown.

chrivers: We'll do more research on this after v1 release :)

Issue #112, opened 2023-05-24
[ENHANCEMENT] Print from Web UI


billyjbryant:
Your issue may already be reported! Please search on the [issue tracker](https://github.com/Ankermgmt/ankermake-m5-protocol/issues) before creating one.

**Describe the solution you'd like**
Provide a way to upload gcode files from the WebUI

**Describe alternatives you've considered**
Uploading through a Slicer or the CLI

**Additional context**

chrivers:
That should be pretty easy, since we already have a http api for print job upload.

I think the main challenge is actually figuring out how to integrate it into the ui.

Maybe we talk about all the upcoming ideas for features on the discord, and see if we can't come up with a plan for an overall ui design to target :)

billyjbryant:
I implemented this in #113 feel free to comment on how I included it :)

chrivers:
Fixed in v1.1 :)

PR #113, opened 2023-05-24
Print from the web interface

billyjbryant:
### Description

This PR adds a new upload button to the home page that allows for uploading gcode files directly from the Web UI

### Preview

image

### Console logs:

```sh
[*] FileTransferService: Requesting start
[*] FileTransferService: Worker started
[*] Going to upload 12729275 bytes as 'wall-honeycomb-224x190size_mk3s__0.20mm_PLA_M5.gcode'
[*] Requesting file transfer..
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] Sending file metadata..
[*] Sending file contents..
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] TOPIC [/phone/maker/AK7ZRG0C35700749/notice]
[*] File upload complete. Requesting print start of job.
[*] Successfully sent print job
[*] FileTransferService: Requesting stop
[*] FileTransferService: Worker stopped
```

### Testing PPPP print-file
To ensure that `pppp print-file` continues to work I tested it as well: 

```sh
billy  …/ankermake-m5-protocol   web-gcode-upload $  ♥ 08:34  ./ankerctl.py pppp print-file -n /mnt/c/Users/billy/OneDrive/Documents/3dPrinterSD/0_SLICED/FDM/wall-honeycomb-224x190size\(mk3s\)_0.20mm_PLA_M5.gcode 
[*] Trying connect to printer Vulcan (USPRAKM-008574-TPUPS) over pppp using ip 192.168.4.145
[*] Established pppp connection
[*] Going to upload 12729275 bytes as 'wall-honeycomb-224x190size_mk3s__0.20mm_PLA_M5.gcode'
[*] Requesting file transfer..
[*] Sending file metadata..
[*] Sending file contents..
100%|████████████████████████████████████████████████████████████████████████████████████████████| 12.1M/12.1M [00:18<00:00, 694kb/s]
[*] File upload complete
```

Fixes: #112


web/util.py
```diff
@@ -24,3 +25,19 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
...
+
+def upload_file_to_printer(app, file, web_upload=False):
+    user_name = request.headers.get("User-Agent", "ankerctl").split(url_for('app_root'))[0]
+
+    with app.svc.borrow("filetransfer") as ft:
+        try:
+            ft.send_file(file, user_name)
+        except ConnectionError as err:
+            log.error(f"Connection error: {err}")
+            raise ConnectionError(err)
```
chrivers, line 38: This pattern is not good. In fact there are 3 problems :sweat_smile:

1) We are creating exception objects where the argument to the exception, is an exception of the same type.
2) Re-raising exceptions in python has an explicit syntax: `raise` (just the keyword, used inside an `except` clause). Otherwise, the backtrace into the actual error site is lost, and the exceptions seem to originate "here". (around python 3.7, I think, the language gained some support for working around this, but it's not kosher)
3) We are only catching these exceptions to log them, and then we have to catch them again when we call `upload_file_to_printer()`, so this doesn't much help us.

web/util.py
```diff
@@ -24,3 +25,19 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
...
+        except Exception as err:
+            log.error(f"Unknown error: {err}")
+            raise Exception(err)
+
+    return {}
```
chrivers, line 43: We're not using the result, so we shouldn't have a `return` here.

web/__init__.py
```diff
@@ -219,6 +219,28 @@ def app_api_ankerctl_server_reload():
...
 
+@app.post("/api/ankerctl/file/upload")
+def app_api_ankerctl_file_upload():
+    if request.method != "POST":
+        return web.util.flash_redirect(url_for('app_root'))
+    if "gcode_file" not in request.files:
+        return web.util.flash_redirect(url_for('app_root'), "No file found", "danger")
+    file = request.files["gcode_file"]
+
+    try:
```
chrivers, line 230: It becoming clear that we are doing a whole lot of manual try-except handling for flask endpoints, and it's only getting worse.

This is not a criticism of your PR, but it's something we need to solve. I'll look into refactoring our handlers, so we can catch exceptions in a more generic way, and return reasonable http messages more automatically :)

static/ankersrv.js
```diff
@@ -97,13 +114,13 @@ $(function () {
...
             name,
             url,
-            badge=null,
-            open=null,
-            close=null,
-            error=null,
-            message=null,
-            binary=false,
-            reconnect=1000,
+            badge = null,
```
chrivers, line 117: There are a lot of seemingly unrelated changes in this file - not all of them not for the better, I have to say. Is this automatic reformatting, or a deliberate change?

I'd really appreciate it if we can avoid mixing semantic and stylistic changes, as it makes the PRs much harder to review. Optimally, general cleanup/refactor/reformatting should go in a separate PR, but at least in independent commits. Please? :smile:

static/ankersrv.js
```diff
@@ -5,20 +5,37 @@ $(function () {
...
+        modalInner.innerText = trigger.dataset.msg;
+        if (trigger.dataset.href) {
+            window.location.href = trigger.dataset.href;
+        }
+    });
+
+    /**
+     * Opens modal if gcode upload file is present
+     */
+    document.getElementById('gcode-upload').addEventListener('click', function(event) {
```
chrivers, line 24: We can use `cash.js` here, to make it easier:

```js
$("#gcode-upload").on("click", function (event) {
...
```

The reason I didn't use it for `popupModal.addEventListener("shown.bs.modal", ...`, is because it doesn't work. Something about `cash.js` and the bootstrap events don't play well together.

static/base.html
```diff
@@ -30,13 +30,16 @@
             <div class="modal-dialog modal-dialog-centered">
                 <div class="modal-content">
                     <div class="modal-body">
-                        <p>Restarting ankerctl...</p>
+                        <p id="modal-inner">Restarting ankerctl...</p>
```
chrivers, line 33: If I understand you correctly, we want to be able to use the modal generically, right?

In that case, shouldn't `modal-inner` be blank here? Or did I misunderstand what we're going for?

static/tabs/setup.html
```diff
@@ -15,16 +15,15 @@
...
-                        </div>
+                        <div class="card-header fs-4">Upload AnkerMake Login File</div>
 
                         <div class="card-body input-group">
                             <input
                                 class="form-control"
                                 type="file"
                                 id="loginFile"
                                 name="login_file"
+                                accept=".json"
```
chrivers, line 26: When testing, I often keep a `login.bak` or similar file on hand. I don't see any reason why we should artificially restrict what the file can be called - we should just send it, and let the backend tell us if it works or not :)

static/ankersrv.js
billyjbryant: Automatic formatting that was unintentional. We designated style guides for python but not for html and javascript files so Prettier stepped in. I'll see if I can revert some of this for the PR.

static/ankersrv.js
billyjbryant: Sure can.

static/base.html
billyjbryant: Yes, theoretically it should be blank, I just left it as is for no real reason at all :D

web/util.py
```diff
@@ -24,3 +25,13 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
...
+
+
+def upload_file_to_printer(app, file, web_upload=False):
+    user_name = request.headers.get("User-Agent", "ankerctl").split(url_for('app_root'))[0]
+
+    with app.svc.borrow("filetransfer") as ft:
+        try:
+            ft.send_file(file, user_name)
+        except:
+            raise
```
chrivers, line 37: Okay, so:

```python3
try
    foo()
except:
    raise
```

is the same as

```python
foo()
```

:smile:


And at this point, the whole function is pretty small (still fine to keep, but maybe it's the `user_name` deduction we should make a helper for?

chrivers:
Also, please retarget to `upcoming/v1.1` :)

web/util.py
billyjbryant: ✅

static/tabs/setup.html
billyjbryant: ✅

static/ankersrv.js
billyjbryant: ✅

static/ankersrv.js
billyjbryant: ✅

web/util.py
billyjbryant: ✅

web/util.py
billyjbryant: The return here is key in telling the function that called the upload that it is completed, otherwise it doesn't redirect. to tell you the upload was successful. I tested this many times with and without `return`, return was necessary. 

✅

static/tabs/home.html
```diff
@@ -15,8 +15,37 @@
                     muted
                     id="player"
                     poster="{{ url_for('static', filename='img/load-screen.svg') }}"
+                ></video>
```
chrivers, line 18: I'd really like to get rid of the last few unrelated changes.

It's not that I'm trying to be difficult, it's because they make merging and rebasing much more difficult.

So for example, this kind of change:

```
></video>
```
vs
```
>
</video>
```

Is particularly problematic. It's almost guaranteed to cause a merge conflict.

Compared to:

```
  <span id="progress" class="position-absolute w-100 fw-bold">
    0%
  </span>
```

vs

```
  <span id="progress" class="position-absolute w-100 fw-bold"> 0% </span>
```

These are a whole number of lines being replaced, so they are less "sharp", so to speak.

Anyway, I'm not doing it to annoy you, I'm doing it to not annoy me :laughing:

static/base.html
```diff
@@ -57,8 +57,7 @@
                     class="btn-close btn-sm btn-close-white"
                     data-bs-dismiss="alert"
                     aria-label="Close"
-                >
-                </button>
```
chrivers, line 61: Please avoid unrelated changes. See below.

billyjbryant:
I played around with putting this on the right bar but it makes the bar too long (IMO). 
image

static/ankersrv.js
```diff
@@ -119,35 +141,30 @@ $(function () {
 
         _open() {
             $(this.badge).removeClass("text-bg-success text-bg-danger").addClass("text-bg-warning");
-            if (this.open)
-                this.open(this.ws);
+            if (this.open) this.open(this.ws);
```
chrivers, line 144: Are all these formatting changes deliberate, or an artifact of vscode?

If you'd like it to look like this, we can talk about it.

...but if it's just the computer having an opinion, please cut it out :stuck_out_tongue:

chrivers:
> I played around with putting this on the right bar but it makes the bar too long (IMO). 

Nice idea though.

We really don't need the file selector most of the time. Maybe we could just have an "upload" button, that then opens a modal with a file selector and instructions? That way it wouldn't take much space at all.

billyjbryant:
> We really don't need the file selector most of the time. Maybe we could just have an "upload" button, that then opens a modal with a file selector and instructions? That way it wouldn't take much space at all.

That sounds like an idea. Maybe we can put it on the top menubar? or where else would you envision seeing it?

static/ankersrv.js
billyjbryant: I'll go through and clean all of the formatting changes up. I also added in some config to `.vscode/settings.json` to force vscode not to auto format anything but what was modified

web/util.py
```diff
@@ -24,3 +25,12 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
         flash(message, category)
 
     return redirect(path)
+
+
+def upload_file_to_printer(app, file):
+    user_name = request.headers.get("User-Agent", "ankerctl").split(url_for('app_root'))[0]
```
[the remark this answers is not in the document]
billyjbryant: That was in your original function, I didn't add it so /shrug :D

web/util.py
```diff
@@ -24,3 +25,12 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
...
+
+    with app.svc.borrow("filetransfer") as ft:
+        ft.send_file(file, user_name)
+
+    return
```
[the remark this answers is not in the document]
billyjbryant: Read my previous comment on the matter, the `return` is warranted and needed because without it, the function that called this from flask never reached the redirect on successful upload and thus the modal stays on the screen and the user is never told whether or not the upload completed.

chrivers:
> > We really don't need the file selector most of the time. Maybe we could just have an "upload" button, that then opens a modal with a file selector and instructions? That way it wouldn't take much space at all.
> 
> That sounds like an idea. Maybe we can put it on the top menubar? or where else would you envision seeing it?

Yeah, that might work - I'm also not quite sure where we could fit it.

How about we just make a new tab for this? Like, completely regular, like the other 3?

That would be super simple, and easy to integrate.

static/ankersrv.js
chrivers: Sounds good :)

web/util.py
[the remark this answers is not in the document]
chrivers: Okay, I had to fire up git blame :stuck_out_tongue: 

We're both to blame. You did a search-replace in commit 1969a29fb3e8099b614bc43c0ae533f24121a4e2, and I didn't catch it in the review.

By sheer coincidence, this happens to work, because `url_for('app_root')` returns `/` in "normal" circumstances for this program.

Here's the relevant part of that diff:

```
-    user_name = request.headers.get("User-Agent", "ankerctl").split("/")[0]
+    user_name = request.headers.get("User-Agent", "ankerctl").split(url_for('app_root'))[0]
```

(it's an unrelated problem, so we don't have to fix it here, but feel free)

web/util.py
chrivers: Sorry, my bad! I mixed up some stuff.

I meant to refer to `upload_file_to_printer()` (which *does* have an unnecessary return)

web/util.py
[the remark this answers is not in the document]
billyjbryant: Haha love git blame :D

billyjbryant:
> How about we just make a new tab for this? Like, completely regular, like the other 3?

Like a completely new tab with just the upload and thats it? Also, got thoughts on how to show the progress of the upload in the UI?

just-trey:
I'd suggest putting it with .the other controls on the right. Then hide it once you start the print

billyjbryant:
> I'd suggest putting it with .the other controls on the right. Then hide it once you start the print

@just-trey the problem with hiding it, is that we have no event that would un-hide it when nothing is printing other than a refresh of the page. This is the same reason that we decided not to hide the progress information when there is nothing printing.

chrivers:
> > How about we just make a new tab for this? Like, completely regular, like the other 3?
> 
> Like a completely new tab with just the upload and thats it? Also, got thoughts on how to show the progress of the upload in the UI?

Yes, that was my idea. I realize that tab is not going to be incredibly full of content, but it'd be nice and simple, at least? :)

As for progress, we have a bit of a challenge.

We can certainly stream the file to ankerctl. We could then provide `upload-to-ankerctl` progress updates over the `ctrl` websocket.

(although uploading to ankerctl is probably not that slow, until the files get quite large)

A bigger problem, is that when transferring from ankerctl to the printer, we have to *start* the message with the complete md5sum and size of the file. So ankerctl must receive the file fully, before we can start sending to the printer.

Once again, we should be able to send a kind of `upload-to-printer` progress updates through the `ctrl` websocket, and have the ui update.

This makes me think we should seriously consider how much we could connect to the upcoming moonraker api implementation, so we don't have to do all this work twice :)

I mean, as long as we have an API, it shouldn't matter too much for our frontend what flavor it is.

chrivers:
As for where we place the feature - I'm not religious about it, but I do agree with you that it's hard to make it fit nicely in the right-hand menu.

If we make a separate tab for it, we could get support for having a print queue, saving old jobs, a history, etc. All that would make sense to have on a "Jobs" tab, I think.

Do you agree?

just-trey:
> 

+1 if we want to go that far with the implementation

static/tabs/home.html
billyjbryant: ✅

billyjbryant:
> If we make a separate tab for it, we could get support for having a print queue, saving old jobs, a history, etc. All that would make sense to have on a "Jobs" tab, I think.

How would we envision a print queue working?

billyjbryant:
This is what I had come up with last night. I can make this a separate tab if we think that is a better flow. 
[screen-capture.webm](https://github.com/Ankermgmt/ankermake-m5-protocol/assets/3013565/24bca27b-9308-468b-91e1-80a788767992)

chrivers:
> This is what I had come up with last night. I can make this a separate tab if we think that is a better flow. [screen-capture.webm](https://github.com/Ankermgmt/ankermake-m5-protocol/assets/3013565/24bca27b-9308-468b-91e1-80a788767992)

I don't mind the idea of a popup, although I personally prefer a separate tab.

But I don't think it makes sense for one of the tab bar buttons to open a popup, while the rest change the active tab. It's just weirdly inconsistent.

If we go that, we should make it not part of the tab list, and make it stand out a bit more:

image


I'd prefer we just make a tab, though. Even though we don't have a lot of content for that tab right now, I think it would make sense to put it there. Especially since we are probably going to implement some job queue features for moonraker anyway.

What do you think?

billyjbryant:
I'm not opposed to a tab, I just think it will look weird, in order to not have the footer suddenly jump up i'm going to have to force height of the container DIV. I'll play around with it tonight. 

I do like the idea of having a leading button that is more separated from the tab menu and a different color to control the popup. Maybe we can go the popup route for the time being then develop a tab around it once we are ready to add other capabilities?

chrivers:
> How about this, I will build both, and let you decide which looks better :) good ole fashioned A-B test :D

Sounds great! I like that :)

billyjbryant:
@chrivers ok, try this on and see how you like it. :)

chrivers:
This looks great!

I've been trying it out locally, and I have to say I prefer the tab style, where we have a tab for printing. I think it makes a lot of sense, and integrates well with the existing ui.

While looking at it, I had an idea that I want to run by you.

The instructions for hooking up PrusaSlicer take an entire tab right now. It's great that we have detailed instructions, but it's something that users very rarely need to do.

When we get instructions for other slicers, we need a way to collapse / navigate to the individual instructions.

So how about we (in a future PR!), make the Print tab (from this PR) show 2 cards:

| Local print   |
|---------------|
| `[file input]` `upload form` |

| PrusaSlicer Integration |
|-------------------------|
| collapsible section     |

The user can click the prusaslicer card, to get the detailed instructions.

That way, we don't need a prominent tab for rarely-used intructions, and we collect all methods of printing into one tab.

That seems like a good solution to me - what do you think?

If you agree, let's work this PR into adding the "Print" tab, and then we can add on the slicer rework in another PR afterwards :)

static/index.html
```diff
@@ -9,23 +9,43 @@
...
+        {% endcall %}
+
+        {% call macro.menu_button(id="print-tab", target="print") %}
+            {{ macro.bi_icon("file-earmark-arrow-up", "Print") }}
         {% endcall %}
     {% endif %}
 
     {% call macro.menu_button(id="setup-tab", target="setup", active=not configure) %}
-        {{ macro.bi_icon("gear-fill", "Setup") }}
+        {{ macro.bi_icon("gear-fill", "Configure") }}
```
chrivers, line 49: We should keep the name as "Setup" :)

web/util.py
```diff
@@ -24,3 +25,10 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
         flash(message, category)
 
     return redirect(path)
+
+
+def upload_file_to_printer(app, file):
+    user_name = request.headers.get("User-Agent", "ankerctl").split('/')[0]
```
chrivers, line 31: Double quotes for `"/"` :)

static/ankersrv.js
```diff
@@ -119,35 +160,30 @@ $(function () {
...
         }
 
         _close() {
             $(this.badge).removeClass("text-bg-warning text-bg-success").addClass("text-bg-danger");
             console.log(`${this.name} close`);
             setTimeout(() => this.connect(), this.reconnect);
-            if (this.close)
```
chrivers, line 130: Can we revert these autoformatting changes as well? Thanks :)

billyjbryant:
> If you agree, let's work this PR into adding the "Print" tab, and then we can add on the slicer rework in another PR afterwards :)

Sounds like a good idea to me. I'll go ahead and re-work the PR.

static/ankersrv.js
billyjbryant: ✅

web/util.py
billyjbryant: ✅

static/index.html
billyjbryant: ✅

billyjbryant:
@chrivers take a look tell me what you think

static/index.html
```diff
@@ -9,19 +9,24 @@
     {% if configure %}
         {% include "tabs/home.html" %}
         {% include "tabs/instructions.html" %}
+        {% include "tabs/print.html" %}
     {% endif %}
 
     {% include "tabs/setup.html" %}
 {% endblock %}
-
+
```
chrivers, line 17: Please remove trailing whitespace

billyjbryant: ✅

static/macro.html
```diff
@@ -6,12 +6,26 @@
     <link rel="stylesheet" href="{{ url_for('static', filename=url) }}" />
 {%- endmacro %}
 
-{% macro menu_button(id, target, active=False) %}
+{% macro trigger_button(id, type, target, toggle="pill") %}
```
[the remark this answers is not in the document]
billyjbryant: ✅

static/base.html
```diff
@@ -19,7 +19,10 @@
                     <img src="{{ url_for('static', filename='img/logo.svg') }}" class="me-2" alt="logo">
                     <span class="text-body-emphasis align-self-baseline fs-4">ankerctl</span>
                 </a>
-                <ul class="nav nav-pills align-items-baseline" id="tab" role="tablist">
+                <ul class="nav nav-pills align-items-baseline px-3" id="printBtn" role="tablist">
+                    {% block print_button %}{% endblock %}
```
chrivers, line 23: Isn't this a remnant of the "print" button we removed?

billyjbryant: Yes, I thought I had removed this. I am away from home, feel free to remove it and push to my branch.

PR #115, opened 2023-05-24
chrivers/webserver-mainsail-rpc

chrivers:
This PR is work-in-progress.

Experimentally, we add support for the Moonraker API, allowing us to connect Mainsail to ankerctl. We can also use this API for printing from PrusaSlicer (and potentially other sources).


web/util.py
```diff
@@ -26,6 +26,10 @@ def flash_redirect(path: str, message: str | None = None, category="info"):
     return redirect(path)
 
 
+def flash_redirect_root(path: str, message: str | None = None, category="info"):
```
billyjbryant, line 29: Why does this function take three parameters if it only uses two?  This seems to be a pointless abstraction of code to make the front-side code "prettier".

chrivers: Oh dear, I made a terrible mistake :sweat_smile: 

The "path" parameter was supposed to be removed - that was the whole point of this utility function! :laughing: 

Thanks for looking things over, I'll fix it :+1:

PR #120, opened 2023-05-28
Adding mechanism for entering an emulated fullscreen from url param

billyjbryant:
### Description

Adding mechanism for entering an emulated fullscreen from url param since you cannot enter native fullscreen without interaction

This lets you browse to `http://127.0.0.1:4470/?fullscreen=1` to get an emulated full-screen experience without having to interact with the page first. 

#### Example: 
image


static/ankersrv.js
```diff
@@ -40,15 +66,46 @@ $(function () {
...
+     * @param {boolean} fullscreen - Whether to set or unset the classes (true to set, false to unset).
+     * @param {boolean} emulate - Whether to emulate the fullscreen mode or not.
+     */
+    function setFullscreenClasses(fullscreen = false, emulate = false) {
+        if (fullscreen) {
+            if (emulate) {
+                $("#vmain").addClass("fullscreen-emulate");
+                $("body").addClass("fullscreen-emulate");
+                $("header").addClass("fullscreen-emulate");
+                $("footer").addClass("fullscreen-emulate");
```
chrivers, line 82: We can simplify this, by using css classes a bit more.

Let's put the `fullscreen-emulate` on these 4 things we are selecting here.

Then, we can make a css class `fullscreen-emulate-active`. The js then becomes

```js
if (emulate)
    $(".fullscreen-emulate").addClass("fullscreen-emulate-active");
```

(and vice versa for teardown)

That way, we can tag any dom element we want to apply special rules for, without changing the js in the future.

static/ankersrv.css
```diff
@@ -117,3 +117,26 @@ code {
     border: 0 none;
     padding: 0;
 }
+
+.fullscreen-emulate {
```
chrivers, line 121: Should this be `#vmain.fullscreen-emulate`?

Otherwise, this applies to `body`, `header`, `footer` and `#vmain`. We hide header and footer, but this makes these rules apply to both `body` and `#vmain`, no?

static/ankersrv.css
```diff
@@ -117,3 +117,26 @@ code {
...
+    position: fixed;
+    width: 90vw;
+    height: 99vh;
+    top: 1%;
+    left: 5%;
```
chrivers, line 126: These "almost the whole way" values seem a little arbitrary. What are we trying to do here?

Why not `100%`?

And why also `max-width` / `max-height`?

static/ankersrv.css
```diff
@@ -117,3 +117,26 @@ code {
...
+    max-height: 99vh;
+}
+
+body.fullscreen-emulate {
+    background-color: rgba(0,0,0,0);
+}
+
+header.fullscreen-emulate,
+footer.fullscreen-emulate {
+    display: none;
```
chrivers, line 141: Maybe we should go the bootstrap way, and have a `.fullscreen-emulate-d-none` class, to mirror the `d-none` class from bootstrap?

Then we could have a generic css rule for `.fullscreen-emulate-d-none-active` here, and apply `.fullscreen-emulate-d-none-active` on `.fullscreen-emulate-d-none` from js? (see below)

static/ankersrv.css
billyjbryant: Because using 100vw/vh caused the content to somehow be larger than 100% of the viewport and thus caused scrollbars and the video to be too big. 

I tweaked around with different settings in a few different viewport sizes and found that these settings produced the same visual experience across all the viewports. 

The max-width and max-height were set to override the @media default css which was restricting the max size to 1322px for some reason.

static/ankersrv.js
```diff
@@ -29,8 +29,34 @@ $(function () {
...
+            case "1":
+                return true;
+            default:
+                return false;
+        }
+    }
 
+    // Get URL parameter for fullscreen and apply it if needed, this emulates fullscreen
+    let urlParams = new URLSearchParams(window.location.search);
+    let fullscreenParam = stringToBoolean(urlParams.get("fullscreen"));
```
chrivers, line 54: This breaks the js if `fullscreen` arg is not given - "string is null" error in `stringToBoolean`

static/ankersrv.css
billyjbryant: I'm not certain how that would improve functionality or the ability to follow what is happening with the change.

chrivers: Just an idea - we'll skip for now :+1:

static/ankersrv.css
billyjbryant: I'm pretty certain it overrides but I can test to see if it is inheriting the un-specified class.

static/ankersrv.js
billyjbryant: ✅

static/ankersrv.css
billyjbryant: Tested and pushed the change :) 

✅

static/ankersrv.js
billyjbryant: ahh, that would be easier :D

billyjbryant: ✅

PR #123, opened 2023-05-31
[Enhancement] Pushes Docker Images to tag :edge and :upcoming on merge to Upcoming/*

billyjbryant:
### Description

Updating the GHA so that we push upcoming/edge images with merges to branches named `Upcoming/*`.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -4,6 +4,9 @@ on:
   push:
     tags:
      - v**
+    branches:
+      - 'upcoming/*'
+      - 'master'
   pull_request:
     branches: [ master ]
      
@@ -48,8 +51,12 @@ jobs:
       with:
         images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
         flavor: |
-          latest=${{ github.event_name == 'push' }}
-        tags: ${{ github.ref_name }}
+          latest=${{ github.ref == 'refs/heads/master' }}
+        tags: |
+          type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
+          type=edge,value=edge,enable=${{ startsWith(github.ref, 'refs/heads/upcoming/') }}
+          type=raw,value=upcoming,enable=${{ startsWith(github.ref, 'refs/heads/upcoming/') }}
+          type=ref,event=tag
         
     - name: Build and push Docker images
       uses: docker/build-push-action@v4.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -48,8 +51,12 @@ jobs:
...
         images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
         flavor: |
-          latest=${{ github.event_name == 'push' }}
-        tags: ${{ github.ref_name }}
+          latest=${{ github.ref == 'refs/heads/master' }}
+        tags: |
+          type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
+          type=edge,value=edge,enable=${{ startsWith(github.ref, 'refs/heads/upcoming/') }}
+          type=raw,value=upcoming,enable=${{ startsWith(github.ref, 'refs/heads/upcoming/') }}
+          type=ref,event=tag
```
chrivers, line 59: What does `type=`, `value=` and `event=` do?

.github/workflows/build-and-publish.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     tags:
      - v**
+    branches:
+      - 'upcoming/*'
+      - 'master'
```
chrivers, line 9: Are these conditions (tags, branches, pull_requests) ANDed or ORed together?

billyjbryant: I'm pretty sure these are treated as ORs.

.github/workflows/build-and-publish.yml
billyjbryant: `type` tells docker the type of image (determines the tag type)
`value` tells docker the value for the tag (`value=edge`) means the tag will be `:edge`

  * raw = tag is set to whatever is set for value, so `type=raw,value=latest` will produce an image with the tag `:latest`
  * edge = Image Tag is set to `:edge`, edge also shows up in the registries website as unstable iirc. 
  * ref = Sets tag to the reference for the case of `type=ref,event=tag` it uses the "tag Github event" as a reference so setting a GH Tag of `v1.1` will set a image tag of `:v1.1`

.github/workflows/build-and-publish.yml
billyjbryant: So i validated this, they are treated as ORs. Anything tagged with a `v<something>`, pushed to `master`, pushed to `upcoming/<something>` or pull requests against `master` will trigger the workflow.

billyjbryant:
@chrivers can we merge this PR so that future Docker releases are tagged so that users can utilize bleeding-edge Docker images without having to build manually?

PR #138, opened 2023-12-13
wss:// support

snoj:
Added ability to use wss for ssl deployments (eg traefik/nginx proxy manager).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 __pycache__
 .DS_Store
 /settings.json
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 __pycache__
 .DS_Store
 /settings.json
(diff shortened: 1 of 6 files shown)


billyjbryant:
Can you provide screenshots as well as documentation on how you configure this for operation?

snoj:
> Can you provide screenshots as well as documentation on how you configure this for operation?

The change doesn't have any configuration needs.  It just allows for a reverse proxy to be able to use certificates/tls.  The original code had hardcoded ws:// for the websocket url in the webui.  This code automatically sets the protocol to wss for the websocket urls if https is being used.

billyjbryant (approved):

billyjbryant:
Can you target the upcoming branch instead of master?

snoj:
I assume you mean "upcoming/v1.1".

Done

PR #139, opened 2023-12-14
Update default IP to wildcard address per Flask Server documentation

clovisd:


billyjbryant:
The default IP is reserved to localhost on purpose, as this prevents accidental exposure of the service and lets you control it via configuration.

billyjbryant:
This does not prevent you from exposing the Flask service to an external service via Docker should you choose to do a Docker port map from host to container.

anselor:
Docker port forwards cannot be used when network mode is set to host.

anselor:
Just a note. My changes in PR #155 allows the bind address to be optionally parameterized and supersedes the changes in this PR.

PR #148, opened 2024-01-29
Add support for fetching configuration via login

treitmayr:
Implemented new command 'config login' to fetch the complete configuration from the server instead of relying on the configuration file created by the AnkerMake Slicer.

Features:
* Options on command line or to be entered interactively by the user.
* Optional region auto-detection by measuring server connection time.
* Support for captcha queries which seem to happen after fetching new credentials multiple times within a short time. Currently the captcha image can only be displayed in a web browser (which might be a problem on headless servers).

Open:
* Update README.
* One or two spots in the code do not look as clean as I would like them to be.

Help output:
```bash
$ ./ankerctl.py config login -h
Usage: ankerctl.py config login [OPTIONS] [REGION (EU/US)] [EMAIL] [PASSWORD]

  Fetch configuration by logging in with provided credentials.

Options:
  -h, --help  Show this message and exit.
```

Usage example:
```bash
$ ./ankerctl.py config login
Please enter your email address: my@email.com
Please enter your password: xxxxx
Please enter your region (EU/US), empty to guess:
[*] Initializing API..
[*] Using region 'EU'
[*] Logging in..
[*] Initializing API..
[*] Requesting profile data..
[*] Requesting printer list..
[*] Requesting pppp keys..
[*] Adding printer [AK75D72D24300601]
[*] Finished import
```


treitmayr:
Unfortunately I just realized, that the IP address in the created printer config is set to an empty string. I will have to check whether it is just not part of the replies from AnkerMake's server, or there is a different issue.

treitmayr:
Added a commit which preserves the printer IP addresses from a valid previous ankerctl configuration if the server replied with empty IP addresses (which it always does for me). So now, you can do `ankerctl.py config login` without loosing these IP addresses.

treitmayr:
The third commit adds a web interface for logging in to AnkerMake's server and fetch the configuration from there. As captchas seem to be requested quite frequently, those are also supported.
This should make it unnecessary to import the slicer login.json file.

Note that it is my first time working with Flask and Bootstrap, and I have tried my best (in the available amount of time) to conform to the existing code style and concepts. For the captcha functionality I had to implement Javascript HTTP requests instead of reloading the page, so this is a newly introduced concept in the ankerctl code. According to my tests it works pretty well for both initial configuration and re-configuring.

Bildschirmfoto vom 2024-02-04 03-12-03-1

sshanky:
I've nearly made it through, but getting this now:
```
[*] Login successful, importing configuration from server..
[*] Initializing API..
[*] Requesting profile data..
[*] Requesting printer list..
[*] Requesting pppp keys..
[!] Config import failed: 'NoneType' object is not iterable
```

P0k3sm0t:
Can you add this info to the main page of fork? Make it a lot easier to figure out after initial setup. I had to look around to find this. Thank you as well. All working great!

treitmayr:
I recommend you switch to the fork [Django1982/ankermake-m5-protocol](https://github.com/Django1982/ankermake-m5-protocol) which incorporates my login support, adds a lot of other features, and is available as a docker image.

PR #145, opened 2024-01-21
Add initial support for the AnkerMake M5C

treitmayr:
This PR fixes #137 and allows to monitor MQTT messages of the AnkerMake M5C on the command line as well as observe the printer status via the Web GUI. It is also possible to send print jobs from your slicer to the printer without using AnkerMake Slicer or Studio (see README.md). The M5's additional hardware is still expected to exist, therefore the Web GUI will still pretend to wait for a camera connection, for example. This will have to be fixed in the future.

Note that I was only able to test with my M5C but not with an M5 for unwanted side-effects.

Eventually, if this patch is confirmed by others to work fine, we should add some related information to the README file.

PR #136, opened 2023-12-07
Fix padding in mqttapi::make_mqtt_pkt

mtoner23:
I tried sending gcode over mqtt and this errored out. even with this fix the printer doesnt respond to the gcode but maybe thats my user error. since we have no clue what this field does i think its fine to do what we were doing before the previous fix; send all zeros.

Issue #158, opened 2024-03-17
[BUG] PPPP service does not stop trying to contact the printer when the last websocket is closed


treitmayr:
**Bug Description**
When the ankerctl web server runs, the printer is not available, and at least one browser is connected to ankerctl, the server tries to contact the printer coninuously, as would be expected so that the PPPP badge can be updated.
However, when the browser (tab) is closed, the server still tries to contact the printer via PPPP. When started as a docker service, this will result in constant write accesses to the log file.

Note: This bug is not related to the number of bugs concerning unsuccessful PPPP connections!

**To Reproduce**
1. Turn off printer
2. Start ankerctl web server
3. Connect with a web browser to the web server
4. Check that the web server starts printing messages
    `Trying connect to printer AnkerMake M5C (EUPRAKM-013657-LNCZN) over pppp using ip <IP>`
5. Close the browser (tab)
6. The ankerctl web server will *not* stop printing the above message

**Expected Behavior**
As soon as there is no websocket connection from any web browser to the PPPP websocket endpoint anymore, ankerctl should stop trying to contact the printer via PPPP.

**Additional Context**
This issue can be observed using the master branch as well as the modified version of PR #145 and [anselor:exiles](https://github.com/anselor/ankermake-m5-protocol/tree/exiles).

**Initial Analysis**
The PPPP service is implemented in a way that it is only "started" when the PPPP could be established. The `ServiceManager.borrow()` and `ServiceManager.stream()` functions called by the websocket endpoint both try to start the PPPP service without employing any timeout meachanism.
Furthermore (and possibly only relevant for the PR #145 version), the PPPP websocket endpoint should check if the websocket connection is still alive inside its loop.

billyjbryant:
Ankerctl does health check connections to the printer as it is an "always on" service. If the service is running, it is communicating to the printer to maintain connection so that when you load the website, it is ready.

Issue #156, opened 2024-03-15
PPPP not working


revil-O:
Printer is found via ankerctl.py pppp lan-search @ 192.168.1.137

[*] Trying connect to printer AnkerMake M5 (EUPRAKM-******-*****) over pppp using ip 
[E] PPPPService: Failed to start worker: [Errno 61] Connection refused. Retrying in 1 second.
Traceback (most recent call last):
  File "/Users/osc/ankermake-m5-protocol/web/lib/service.py", line 103, in _attempt_start
    self.worker_start()
  File "/Users/osc/ankermake-m5-protocol/web/service/pppp.py", line 52, in worker_start
    msg = api.recv(timeout=(deadline - datetime.now()).total_seconds())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/osc/ankermake-m5-protocol/libflagship/ppppapi.py", line 351, in recv
    data, self.addr = self.sock.recvfrom(4096)
                      ^^^^^^^^^^^^^^^^^^^^^^^^
ConnectionRefusedError: [Errno 61] Connection refused

<img width="966" alt="Bildschirmfoto 2024-03-15 um 14 16 06" src="https://github.com/Ankermgmt/ankermake-m5-protocol/assets/3357909/5dc2ce8b-814e-4833-91b7-52c71b10d775">

OS: MacOSX Sonoma 14.2.1
Python: 3.11
Ankermake M5 Fw: 3.2.17_3.0.94

PrusaSlicer is obviously uploading files ... but printing never starts

following information, repeated every 3 seconds , is shown:

[*] Trying connect to printer AnkerMake M5 (EUPRAKM-******-*****) over pppp using ip 
[*] TOPIC [/phone/maker/AK7Z**************/notice]

If you need additional informations, don't hesitate to ask me for.

Regards revil.O

ngiaimo:
I am getting similar but I don't get an error just keeps saying trying to connect over pppp. The IP it gives is what I can see in my router as the assigned IP of the printer and my computer is on the same subnet.

treitmayr:
There usually seem to be two possible issues resulting in the above error:
[...]

PR #160, opened 2024-05-23
Exiles on 1.1

anselor:
Aggregation of many PRs

Integrates the following PRs

#123 
#136 
#138 
#145 
#148 
#150 
#155 

#157 

Supersedes #139


static/ankersrv.js
```diff
@@ -355,18 +426,31 @@ $(function () {
...
     sockets.ctrl = new AutoWebSocket({
         name: "Control socket",
-        url: `ws://${location.host}/ws/ctrl`,
+        url: `${location.protocol.replace('http','ws')}${location.host}/ws/ctrl`,
         badge: "#badge-ctrl",
     });
 
+    sockets.pppp_state = new AutoWebSocket({
+        name: "PPPP socket",
+        url: `ws://${location.host}/ws/pppp-state`,
```
sondregronas: Found an error in the pppp_state that would prevent PPPP connection over https:

```suggestion
        url: `${location.protocol.replace('http','ws')}${location.host}/ws/pppp-state`,
```

anselor: @treitmayr This was from your changes in this PR:
https://github.com/Ankermgmt/ankermake-m5-protocol/pull/145/commits/1089b493d13e6c3b6abbe4a9168972407702e8c4

treitmayr: My changes were done in a consistent way in the existing code base at the time of the commit. However, the merge of [d6c727b](https://github.com/Ankermgmt/ankermake-m5-protocol/pull/160/commits/d6c727b378140d58395d50f12482fec53864b1d3) introduced wss support, again on the code base from back then. This is the explanation for the current state.
However, I agree with @sondregronas to change the affected line in the proposed way.

sondregronas:
Just commenting here to say that I've opened a PR to this PR with some more changes - let me know if you want me to switch targets to the main repo. PR: https://github.com/anselor/ankermake-m5-protocol/pull/4

### Changes within that PR to this one

- updating `pycryptodomex`, `flask`, `flask-sock` to the latest version + adding `python-dotenv` to dependencies
- addresses the change mentioned above in `ankersrv.js`
- restart the pppp api worker (by calling `worker_start()`, which restarts the api state) on pppp failure (see notes below)
- start the videoqueue service on `/video` if it isn't running, allowing webrtc access via go2rtc (see #162)
- bugfix: allow webserver access with missing credentials (webserver couldn't start without any camera present in cfg)
- `/api/ankerctl/status` endpoint, displays a JSON of every services state and the `app_api_version` dict

### Note regarding pppp:

With this version, pppp should not crash (at least not as often). I've noticed before it would stop working after 2-3 hours of uptime, (but that could also be my really old xeon failing). Instead of crashing it will "flicker" for lack of better words, where sending gcode will _sometimes_ work, and the videostream will have corrupted B/P-frames that only get worse over time. Reloading services are to no avail. Restarting the entire application every 2 hours does circumvent this as far as I can tell.

I tried getting the webserver to run with a different WSGI server, but had no luck. I only tested `gevent` or `eventlet`. I wanted to try gunicorn but from my understanding it would require some refactoring to how the cfg is loaded so I stopped trying.

### All of these prs combined _should_ resolve (at least to my limited understanding):

#163, #161, #158, partially #156 (PPPP will _work_ but stops being reliable), #154, #151, #147, #127 (might be unrelated?), #112

Might be missing a few issues as well that I haven't encountered myself.

